PR #121, opened 2015-01-07
Local development and CI via halcyon

pbrisbin:
This PR implements a workflow where local development and CI testing also uses Halcyon (deployments already use Halcyon via the Haskell on Heroku buildpack). The benefits are consistent, reproducible builds and artifact caching between users, tests, and deployments.

This PR has been successfully tested on Arch, Travis, and CircleCI. We'd like to see an OSX user test it before merging, but may merge anyway if too much time passes.

To try out this branch:

- Remove any existing sandbox (`cabal sandbox delete`)
- Remove any existing build artifacts (`cabal clean`)
- Remove any existing `.env` file (`mv .env env.bak`)
- Remove any existing `~/.ghc` (`mv ~/.ghc ~/ghc.bak`)
- Checkout this branch (`git checkout pb-travis-halcyon`)
- Follow the instructions in the README


pbrisbin:
Checkout the duration difference when I added a base dependency (`yesod-bin`) vs just running the build:



master runs ~20 minutes every time.

pbrisbin:
@jferris if you're willing to test (it requires blowing away any current sandbox), I'd love it if you could try the following:

```
% cabal sandbox delete
% rm -rf dist
```

```
% git fetch && git checkout pb-travis-halcyon
% ./bin/setup
```

This should not build anything, it should re-use all the artifacts from my build, it should be fast. Note that it will use `sudo` to create a user-writable `/app` on your system.

Once done, you should be able to:

```
% source bin/halcyon-env
% yesod test
% yesod devel
% # etc...
```

As you would normally work under this new approach.

Let me know!

bin/halcyon-build
```diff
@@ -0,0 +1,29 @@
...
+  git clone https://github.com/mietek/halcyon.git /app/halcyon
+fi
+
+# Note --no-clean-private-storage ensures developers don't need to rebuild their
+# stuff if another developer builds other versions for their own use (by
+# default, halcyon keeps only one version per platform at a time).
+/app/halcyon/halcyon build \
+  --no-clean-private-storage --cabal-version=1.22.0.0 "$@"
+
+ln -sf /app/sandbox/.halcyon-sandbox.config cabal.sandbox.config
```
mietek: **Note:**  https://github.com/mietek/halcyon/commit/8d6bd08eed129ccb3e81bdd562cf54cb04ab68b5 changed `.halcyon-sandbox.config` to `cabal.sandbox.config`.

pbrisbin: Thanks!

.halcyon/sandbox-extra-apps
```diff
@@ -0,0 +1,3 @@
+alex-3.1.3
+happy-1.19.4
+yesod-bin-1.4.3.1
```
mietek: You should also add [constraints for the sandbox extra apps](https://halcyon.sh/reference/#halcyon_sandbox_extra_apps_constraints).  For example:
- https://github.com/mietek/hl/blob/master/.halcyon/sandbox-extra-apps-constraints/alex-3.1.4.constraints
- https://github.com/mietek/hl/blob/master/.halcyon/sandbox-extra-apps-constraints/happy-1.19.5.constraints

pbrisbin: Would it not make sense to use the overall constraints when installing these packages?

mietek: No, because the sets of dependencies for your app and for each extra app may be completely disjoints.

When Halcyon installs extra apps, it makes sure none of their dependencies contaminate your app’s sandbox — in contrast to what happens when you just use `cabal install`.

pbrisbin: It's nice that there's no contamination. In the case that the extra apps and my app work across the same set of dependencies (which has been the case so far but may not be likely in general), it'd be nice to only have to manage one file. The ones you liked are small though, so oh well.

mietek: It may not be immediately obvious that Halcyon calls itself recursively in order to install extra apps.

For example, when you specify `alex-3.1.3` as a sandbox extra app, somewhere along the process a command similar to `halcyon install alex-3.1.3 --prefix=/app/sandbox` gets executed.

mietek:
Is there anything I can do to help get this merged?

pbrisbin:
Thanks for the offer @mietek!

Last Friday we tried to get an OSX user to test this. That's the only blocker. He ran into an error in bashmenot about not having gdate, gsort, etc. We haven't started digging yet.

@tonyd256 can you provide details here about what you ran into? Maybe we can get it sorted out with @mietek 's help.

I plan to spend on some time on this tomorrow, so maybe we can all sync up then.

mietek:
The only thing missing on OS X is support for automatically installing native OS packages (https://github.com/mietek/halcyon/issues/16).

@tonyd256 Your issue should be resolved with `brew install coreutils`.

pbrisbin:
In case it wasn't clear: Carnival is using halcyon (via haskell-on-heroku), albeit at an older version, already on master. This PR is just to start using halcyon also for local development and CI.

> Your issue should be resolved with brew install coreutils.

I didn't realize the gnu-tools were required to use halcyon on OSX. That explains it.

mietek:
> I didn't realize the gnu-tools were required to use halcyon on OSX. That explains it.

Right.  I’ll make this clear in the Halcyon documentation once I announce OS X support.  For now, it’s only mentioned on the [_bashmenot_ front page](https://bashmenot.mietek.io/#dependencies), and in the [_bashmenot_ reference](https://bashmenot.mietek.io/reference/#date-formatting-module).

bin/halcyon-build
pbrisbin: To confirm something: this file goes into the cached archives right? That means this line will continue to work for existing archives but fail on any new builds? I guess the solution is to check for both while we're in transition.

mietek: Halcyon already rejects and automatically rebuilds sandbox archives which do not contain the file with the new name.

pbrisbin: That's odd. I just ran this script and it worked fine. It used a cached build, got this file and the `ln` succeeded.

mietek: Can you gist the entire log, please?

pbrisbin: Here you go http://sprunge.us/jBeN

Sorry about the escapes.

mietek: OK, and `ls -l /app/sandbox`?

pbrisbin: http://sprunge.us/AAKP?

mietek: I’m surprised you have a symlink in `/app/sandbox`:
```
/app/sandbox/cabal.sandbox.config -> /app/sandbox/.halcyon-sandbox.config
```

I’d be expecting only the actual config in `/app/sandbox`, and the symlink in `/app`:
```
/app/cabal.sandbox.config -> /app/sandbox/.halcyon-sandbox.config
```

Can you list the contents of your sandbox directory archive?
```
tar -tf /var/tmp/halcyon-cache/halcyon-sandbox-fd49f80.701b05b-carnival-0.0.0.tar.gz
```

pbrisbin: Sure there: http://sprunge.us/DDKb?

mietek: I’ve searched the entire change history for `.halcyon-sandbox.config`, and at no point in time was Halcyon creating a symlink in `/app/sandbox`.  I don’t know how it got into the sandbox directory archive.  Were you doing any manual symlinking?

pbrisbin: Actually, maybe. I remember you giving me the instruction about symlinking the cabal config in IRC, and I misread it the first time, possibly creating the symlink in `/app/sandbox`.

mietek: OK. In any case, you can always force Halcyon to rebuild the sandbox from scratch by using the [`HALCYON_SANDBOX_REBUILD`](https://halcyon.sh/reference/#halcyon_sandbox_rebuild) option.

pbrisbin:
> Is there anything I can do to help get this merged?

FWIW, I'm a little nervous about moving 100% to this setup with https://github.com/mietek/halcyon/issues/5 still open.

pbrisbin:
@jferris thoughts on trying to get an OSX user to test on Friday, but merging then regardless? I think it's time to bite the bullet and deal with whatever fallout is caused by having this on master.

jferris:
Yeah, I sort of think we should just merge this, and then use new pull request to fix any issues we discover with it.

pbrisbin:
I agree.

Travis is all of sudden having trouble with coffee-script. Once I can address that and squash this all down, I'll merge (but feel free move forward with any of that if you want).

pbrisbin:
I've addressed the coffee-script issue, squashed, and rewrote the commit message. I think this is ready to go.

README.md
```diff
@@ -10,91 +10,84 @@ Carnival is maintained and funded by [thoughtbot, inc](http://thoughtbot.com/com
...
+Before doing anything:
 
 ```
-$ ./bin/deploy
-```
-
-The production app can be passed as an argument if desired:
+$ source <(/app/halcyon/halcyon paths)
 
+# POSIX version
```
jferris: Do you think it will be clear to everyone that they don't need to run both of these commands?

pbrisbin: Not sure but I'll clarify anyway

jferris:
This looks good to go. Merge merge merge!

mietek:
:joy:

PR #122, opened 2015-01-09
Abort if Halcyon ENV vars are missing

jferris:
* Prevents accidentally rebuilding all dependencies
* Also abort before printing "Setup complete"

### DIFF
--- a/bin/halcyon-build
+++ b/bin/halcyon-build
@@ -6,6 +6,13 @@ set -e
 [ -d /app/halcyon ] || \
   git clone https://github.com/mietek/halcyon.git /app/halcyon
 
+if [ -z "$HALCYON_AWS_ACCESS_KEY_ID" ] || \
+  [ -z "$HALCYON_AWS_SECRET_ACCESS_KEY" ] || \
+  [ -z "$HALCYON_S3_BUCKET" ]; then
+  echo "Make sure Halcyon environment variables are set." >&2
+  exit 1
+fi
+
 export HALCYON_NO_SELF_UPDATE=1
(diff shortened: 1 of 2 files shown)


bin/halcyon-build
```diff
@@ -6,6 +6,13 @@ set -e
 [ -d /app/halcyon ] || \
   git clone https://github.com/mietek/halcyon.git /app/halcyon
 
+if [ -z "$HALCYON_AWS_ACCESS_KEY_ID" ] || \
+  [ -z "HALCYON_AWS_SECRET_ACCESS_KEY" ] || \
+  [ -z "HALCYON_S3_BUCKET" ]; then
```
pbrisbin: Missing some `$`'s here.

bin/halcyon-build
```diff
@@ -6,6 +6,13 @@ set -e
 [ -d /app/halcyon ] || \
   git clone https://github.com/mietek/halcyon.git /app/halcyon
 
+if [ -z "$HALCYON_AWS_ACCESS_KEY_ID" ] || \
+  [ -z "$HALCYON_AWS_SECRET_ACCESS_KEY" ] || \
+  [ -z "$HALCYON_S3_BUCKET" ]; then
+  printf "Make sure Halcyon environment variables are set.\n" >&2
```
pbrisbin: We waffled on the `printf` vs `echo` thing in Guides. We landed on `echo` for simple cases. Can you use it here for consistency across the script? (note no trailing `\n` is needed with `echo` too).

pbrisbin:
Minor thing -- LGTM. Feel free to commit right to that other branch.

PR #124, opened 2015-01-23
Use prefix to make ident unique by OAuth2 provider

jferris:
We want to replace Upcase with GitHub for authentication. We need a prefix
before identifiers to ensure there are no collisions between Upcase and GitHub
users.

### DIFF
--- a/Foundation.hs
+++ b/Foundation.hs
@@ -131,7 +131,7 @@ instance YesodAuth App where
     logoutDest _ = SessionR
 
     getAuthId creds = runDB $ do
-        muser <- getBy $ UniqueUser $ credsIdent creds
+        muser <- getBy $ UniqueUser (credsPlugin creds) (credsIdent creds)
 
         let newUser = buildUser creds
 
@@ -148,11 +148,12 @@ instance YesodAuth App where
     authHttpManager = httpManager
 
 buildUser :: Creds m -> Maybe User
-buildUser (Creds _ csId csExtra) =
+buildUser (Creds csPlugin csIdent csExtra) =
     User <$> lookup "first_name" csExtra
          <*> lookup "last_name" csExtra
          <*> lookup "email" csExtra
-         <*> pure csId
+         <*> pure csPlugin
+         <*> pure csIdent
 
 replaceUser :: UserId -> Maybe User -> YesodDB App UserId
 replaceUser uid (Just u) = replace uid u >> return uid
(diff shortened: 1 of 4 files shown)


Foundation.hs
```diff
@@ -148,11 +148,12 @@ instance YesodAuth App where
     authHttpManager = httpManager
 
 buildUser :: Creds m -> Maybe User
-buildUser (Creds _ csId csExtra) =
+buildUser (Creds csIdent csPlugin csExtra) =
```
pbrisbin: These are backwards according to http://hackage.haskell.org/package/yesod-auth-1.4.1.3/docs/Yesod-Auth.html#v:Creds

config/models
```diff
@@ -3,7 +3,8 @@ User
     lastName Text
     email Text
     ident Text
-    UniqueUser ident
+    plugin Text
+    UniqueUser ident plugin
```
pbrisbin: I would do plugin then ident. It reads better to me and is the order of the `Creds` datatype.

migrations/add-plugin-to-user-ident.sql
```diff
@@ -0,0 +1,5 @@
+ALTER TABLE "user" ADD COLUMN plugin varchar;
+UPDATE "user" SET plugin = 'learn';
+ALTER TABLE "user" ALTER COLUMN plugin SET NOT NULL;
+ALTER TABLE "user" DROP CONSTRAINT unique_user;
+ALTER TABLE "user" ADD CONSTRAINT "unique_user" UNIQUE(ident, plugin);
```
pbrisbin: This is probably fine here, but restarting the app will auto-add this index. It might even drop the existing one first (since the name is the same) but that I'd have to test.

Foundation.hs
jferris: Good catch. Too bad the `plugin` field can't be a different type.

migrations/add-plugin-to-user-ident.sql
jferris: It looks like it does drop and create it. I'm still getting used to the idea of auto-migrating some fields.

pbrisbin:
LGTM -- I'm surprised you didn't have to update any tests. I guess we don't construct users in them.

jferris:
> I'm surprised you didn't have to update any tests. I guess we don't construct users in them.

Whoops, looks like I do. I forgot `yesod devel` doesn't run them. Fixing those now.

pbrisbin:
Good to merge on Green.

PR #125, opened 2015-01-23
Merge first and last names into name

jferris:
GitHub auth (among other providers) only provides a "name" and not distinct
first and last names. We only ever use the full name, so we can safely merge
these into a single name when we authenticate into providers which provide
both.

### DIFF
--- a/Foundation.hs
+++ b/Foundation.hs
@@ -12,6 +12,8 @@ import Network.HTTP.Conduit (Manager)
 import qualified Settings
 import Settings.Development (development)
 import qualified Database.Persist
+import Data.Monoid ((<>))
+import Data.Text (Text)
 import Database.Persist.Sql (SqlBackend)
 import Settings.StaticFiles
 import Settings (widgetFile, Extra (..), LearnOAuthKeys (..))
@@ -149,11 +151,16 @@ instance YesodAuth App where
 
 buildUser :: Creds m -> Maybe User
 buildUser (Creds csPlugin csIdent csExtra) =
-    User <$> lookup "first_name" csExtra
-         <*> lookup "last_name" csExtra
+    User <$> buildName csExtra
          <*> lookup "email" csExtra
          <*> pure csPlugin
          <*> pure csIdent
+
+buildName :: [(Text, Text)] -> Maybe Text
+buildName csExtra = do
+    firstName <- lookup "first_name" csExtra
+    lastName <- lookup "last_name" csExtra
+    return $ firstName <> " " <> lastName
 
 replaceUser :: UserId -> Maybe User -> YesodDB App UserId
 replaceUser uid (Just u) = replace uid u >> return uid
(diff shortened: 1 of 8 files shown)


Foundation.hs
```diff
@@ -149,12 +151,19 @@ instance YesodAuth App where
...
 buildUser (Creds csPlugin csIdent csExtra) =
-    User <$> lookup "first_name" csExtra
-         <*> lookup "last_name" csExtra
+    User <$> buildName csExtra
          <*> lookup "email" csExtra
          <*> pure csPlugin
          <*> pure csIdent
 
+buildName :: [(Text, Text)] -> Maybe Text
+buildName csExtra = T.intercalate " " <$> parts
```
jferris: I think maybe this is more readable with `do` syntax:

``` haskell
buildName :: [(Text, Text)] -> Maybe Text
buildName csExtra = do
    firstName <- lookup "first_name" csExtra
    lastName <- lookup "last_name" csExtra
    return $ firstName <> " " <> lastName
```

Thoughts?

migrations/merge-user-first-and-last-name-into-name.sql
```diff
@@ -0,0 +1,5 @@
+ALTER TABLE "user" ADD COLUMN name varchar;
+UPDATE "user" SET name = first_name || ' ' || last_name;
```
pbrisbin: `||` is string concatenation? Weird.

tests/TestHelpers/DB.hs
```diff
@@ -36,8 +36,7 @@ runDB query = do
 createUser :: Text -> Example (Entity User)
 createUser ident = do
     insertEntity User
-        { userFirstName = "John" <> ident
-        , userLastName  = "Smith"
+        { userName      = "John " <> ident <> " Smith"
```
pbrisbin: Does it break anything to do `"John Smith (" <> ident <> ")"`? I think that'd be better overall, but I understand if you want to keep the changes here to a minimum.

Foundation.hs
```diff
@@ -149,12 +151,17 @@ instance YesodAuth App where
...
+    User <$> buildName csExtra
          <*> lookup "email" csExtra
          <*> pure csPlugin
          <*> pure csIdent
 
+buildName :: [(Text, Text)] -> Maybe Text
+buildName csExtra = do
+    firstName <- lookup "first_name" csExtra
+    lastName <- lookup "last_name" csExtra
+    return $ firstName <> " " <> lastName
```
pbrisbin: Thoughts on moving `buildUser`, `buildName`, and `repalceUser` into `Model.User`?

pbrisbin:
Minor comments -- LGTM

migrations/merge-user-first-and-last-name-into-name.sql
jferris: It is string concatenation, and it is weird.

tests/TestHelpers/DB.hs
jferris: It's not much different to do it in the parens way, and I don't really have a preference. I'll update it to use that style.

Foundation.hs
jferris: That seems logical to me, but I think it means a larger restructuring than we can perform here:

```
Module imports form a cycle:
         module ‘Import’ (Import.hs)
        imports ‘Foundation’ (Foundation.hs)
  which imports ‘Model.User’ (./Model/User.hs)
  which imports ‘Import’ (Import.hs)
```

pbrisbin: Oh right. Only Model can be imported into Foundation. Ignore me.

PR #126, opened 2015-01-23
Use GitHub for authentication

jferris:
Most people don't have an Upcase account, and people who sign up for one via
the blog rarely become Upcase members in any other sense. It makes more sense
both for Upcase and Carnival to use a more common form of authentication.


pbrisbin:
LGTM. Nice to see how relatively easy this switch-over was from a code standpoint.

PR #127, opened 2015-01-24
Update heroku-buildpack-nodejs, Node, NPM

jferris:
* We have no modifications in our fork
* Pointing to a tag makes it easy to update
* The versions we were using were very out of date
* NPM stopped hosting the versions we wanted


pbrisbin:
LGTM

PR #128, opened 2015-01-30
Update GitHub OAuth2 instructions

jferris:
* Most users do not have access to our OAuth2 applications
* Any valid client ID and secret will work for development


pbrisbin:
LGTM

PR #130, opened 2015-02-06
Make bin/setup work for non-thoughtbot users

pbrisbin:
Further details in the commit messages.

### DIFF
--- a/bin/halcyon-build
+++ b/bin/halcyon-build
@@ -9,9 +9,21 @@ private_storage_setup() {
 
 [ -f ./.env ] && source ./.env
 
-if ! private_storage_setup; then
-  echo 'You must set the Halcyon variables for AWS credentials' >&2
-  echo 'https://halcyon.sh/reference/#private-storage-options' >&2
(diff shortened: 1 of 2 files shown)


bin/halcyon-build
```diff
@@ -10,9 +10,17 @@ private_storage_setup() {
 [ -f ./.env ] && source ./.env
 
 if ! private_storage_setup; then
-  echo 'You must set the Halcyon variables for AWS credentials' >&2
-  echo 'https://halcyon.sh/reference/#private-storage-options' >&2
-  exit 1
+  cat >&2 <<EOF
```
jferris: It seems likely to me that people will miss this message, as it will print just before Halcyon starts building everything.

What do you think about keeping this as an `exit 1`, but having an environment variable you can set to ignore it?

pbrisbin: Yeah, I guess. I did consider that option, but ultimately decided against. I don't feel strongly though, so I'll go ahead with your suggestion.

pbrisbin:
Ready for a second look

jferris:
Looks good.

PR #131, opened 2015-02-10
Use upstream Haskell on Heroku buildpack

pbrisbin:
Upstream has significant updates and fixes compared to our out of date fork
and no longer requires the PATH workaround motivating our fork in the first
place.

I've pegged it at the current HEAD revision to ensure consistency. According
to the developer, he won't be consistently tagging releases and v1.0.0 is
lacking some critical fixes with regard to slug size.

Notes:

- Once merged, we should delete our fork from GitHub
- Deployment will require a force-push since I've been testing this branch
  directly on staging


jferris:
Looks good to me.

PR #134, opened 2015-02-19
Install coffee-script as Halcyon sandbox-extra-os-packages

pbrisbin:
On supported platforms (Heroku, Travis), Halcyon now installs node and
coffee-script into /app/sandbox. This means we no longer need to commit
node-related code or use the multi buildpack. On unsupported platforms
(Arch, OS X), a coffee-script installation is still assumed.

https://halcyon.sh/reference/#halcyon_sandbox_extra_os_packages


.halcyon/sandbox-pre-build-hook
```diff
@@ -0,0 +1,2 @@
+#!/bin/sh
+sed -i 's|^#!/usr/bin/nodejs|#!/usr/bin/env nodejs|' /app/sandbox/usr/bin/coffee
```
jferris: Ha.

jferris:
This looks good to me. Awesome that we no longer need to bundle the node modules.

PR #135, opened 2015-02-20
Ensure pre-build hook succeeds on non-ubuntu platforms

pbrisbin:
We only install the coffee-script extra-os-package on certain platforms. On
platforms where we don't, the pre-build script was failing the build because
/app/sandbox/usr/bin/coffee didn't exit.


pbrisbin:
Ping @jferris

jferris:
Should we ensure `coffee` is on the `PATH` for cases where it's not in the sandbox?

pbrisbin:
When thinking about developers as end-users, I personally prefer the transparency of letting real errors happen vs adding code to guess/check/inform. Sure, there's not much effort to adding a check in this case, but there's also not much value (IMO).

The real error would like this:

```
coffee: ... (not such file or directory)
```

Which could be improved, but I think is informative *enough*.

jferris:
Okay; I'd be fine with merging this as-is and seeing what happens when others go to set it up.

Issue #105, opened 2014-10-07
Normalize comment data


pbrisbin:
Having all data in a single comments table will be difficult to scale as we add features. The general sense is that breaking out a few more tables would be a Good Idea and we should do it sooner rather than later before we accrue more data and build more features on top of the current model layer.

Proposed structure:

```
Article
  path (default: location.path)
  title (default: document.title)
  author
  belongs_to Site

Thread
  name (generated from paragraph content from client-side)
  belongs_to Article

Comment
  body
  created
  belongs_to Thread
  belongs_to User
```

PR #136, opened 2015-03-02
Change thoughtbot references to use https

jferris:
* We now support https for the blog
* We have an SSL endpoint for this app on Heroku
* We have a wildcard cert for thoughtbot.com subdomains
* We need to serve https references to avoid unsafe warnings


mike-burns:
Can some of these values be pulled out into a setting or, in general, a shared variable?

pbrisbin:
I'd rather save that to do as part #105 (e.g. extract a `Site` model and make thoughtbot just the first record).

jferris:
I agree with @pbrisbin about the `Site` model.

jferris:
Good to merge?

pbrisbin:
I think so.

PR #137, opened 2015-03-02
Configure gravatar to use https

jferris:
* We're serving the blog over https
* We get mixed content warnings for http images


pbrisbin:
LGTM

PR #139, opened 2015-03-02
Update test suite for newer practices

pbrisbin:
Big improvements:

- No more ydescribe/yit
- No more manually clearing the database between specs
- Spec files can be run in isolation

Smaller improvements:

- Use (Yesod/Haskell) conventional test/ directory (not tests/)
- Use (thoughtbot) conventional Foo/Bar/Spec.hs (not FooTest.hs)
- Improve organization of describe blocks in CommentSpec
- Improve valueEquals (compare Value/Value, not String/String)

There is a lot of renaming, but I tried to isolate those commits if you're
interested in seeing actual differences.


.ghci
```diff
@@ -1,4 +1,5 @@
 :set -i.:config:dist/build/autogen
+:set -DDEVELOPMENT
```
pbrisbin: Now that we can `ghci FooSpec.hs` to run individual specs, this is important to prevent an exception if `getEnv` looks for `DATABASE_URL` in non-development.

jferris:
> Use (thoughtbot) conventional `Foo/Bar/Spec.hs` (not `FooTest.hs`)

Is this also a community convention, or are we off the beaten path with this?

pbrisbin:
I hedged when I wrote that note, but on second look, it is a community convention.

I was confused because the yesod scaffold used to use HUnit and name files `FooTest.hs` (that's how we got here). Now that it's using Hspec though, it names files `FooSpec.hs`[1][] -- I didn't realize that was an HUnit/Hspec difference and thought maybe we were breaking with the scaffold for some reason in our other apps. It's nice that we aren't.

[1]: https://github.com/yesodweb/yesod-scaffold/tree/postgres/test/Handler

pbrisbin:
Travis failure is that lack of `cabal update` and not finding newer gravatar. Thoughts on having Travis always run `cabal update` as part of building?

test/TestHelper.hs
```diff
@@ -0,0 +1,81 @@
...
+-- Test helpers
+import TestHelpers.DB as X
+import TestHelpers.Request as X
+import TestHelpers.Assertions as X
+import TestHelpers.Auth as X
+
+import qualified Data.Text as T
+
+withApp :: SpecWith App -> Spec
+withApp = before $ do
```
jferris: I think this works:

``` haskell
withApp = before $ wipeDB *> makeFoundation <<= testConfig
```

Any thoughts on that? I feel like `<*`/`*>` do a good job of encapsulating the "do a side effect based on this and then pass through" pattern.

test/TestHelper.hs
```diff
@@ -0,0 +1,81 @@
...
+    wipeDB foundation
+    return foundation
+
+testConfig :: IO (AppConfig DefaultEnv Extra)
+testConfig =
+    Yesod.Default.Config.loadConfig
+        $ (configSettings Testing)
+        { csParseExtra = parseExtra }
+
+wipeDB :: App -> IO ()
```
jferris: This is nice. Thoughts on submitting this to Persistent or Yesod?

test/TestHelper.hs
pbrisbin: Should be try to maintain some level of formatting parity with the current scaffold?

https://github.com/yesodweb/yesod-scaffold/blob/postgres/test/TestImport.hs

I'm leaning no. Partly because we have to change some things, and partly because I'm not sure the value.

test/TestHelper.hs
```diff
@@ -0,0 +1,81 @@
...
+wipeDB app = do
```
jferris: Do you need this `do`?

test/TestHelper.hs
```diff
@@ -0,0 +1,81 @@
...
+    runDBWithApp app $ do
```
jferris: Is the return type of this actually `IO ()`? I'm sort of surprised it isn't some db-related type.

jferris:
Does Yesod have something to replace the watchr setup now?

pbrisbin:
> Does Yesod have something to replace the watchr setup now?

No, I just thought I was the only person who every used it. And I no longer do, since I can `ghci` some spec now.

test/TestHelper.hs
jferris: Yeah, I think if we're changing something, it makes sense to format it our own way.

jferris:
Travis looks busted because of my gravatar change, but the code here looks good to me.

test/TestHelper.hs
pbrisbin: I can't get your expression to type-check :/ going to leave as-is for now.

test/TestHelper.hs
pbrisbin: Having this run in `IO` is the only way it could be used in `Spec` (via `liftIO`).

pbrisbin:
Travis will now always run `cabal update` -- will merge on Green.

PR #140, opened 2015-03-03
Use interpolated, type-safe URLs in coffee-script

pbrisbin:
- Removes CarnivalOptions.host
- Removes client-side URL building
- Ensures that URL changes will not compile without updating the javascript;
  when we change endpoints for #105, we can't forget coffee-script changes

### DIFF
--- a/templates/embed/Article.coffee
+++ b/templates/embed/Article.coffee
@@ -28,7 +28,7 @@ class Article
         Carnival.removeClass(@element, 'commenting')
 
   fetchComments: ->
-    Carnival.get('/comments?article=' + @id, (data) =>
+    Carnival.get('@{CommentsR}?article=' + @id, (data) =>
       @commentData = data.comments
       @createBlocks()
       @insertBlocksIntoDom()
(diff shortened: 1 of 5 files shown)


templates/embed/Article.coffee
```diff
@@ -28,7 +28,7 @@ class Article
         Carnival.removeClass(@element, 'commenting')
 
   fetchComments: ->
-    Carnival.get('/comments?article=' + @id, (data) =>
+    Carnival.get('@{CommentsR}?article=' + @id, (data) =>
```
jferris: Just checking: will this include the full schema/host by default?

jferris:
This seems like a good change. I had one question to make sure I understand why it works.

templates/embed/Article.coffee
pbrisbin: Yes, because of [this](https://github.com/thoughtbot/carnival/blob/master/Foundation.hs#L55), it will use [this](https://github.com/thoughtbot/carnival/blob/master/config/settings.yml#L4), [this](https://github.com/thoughtbot/carnival/blob/master/config/settings.yml#L15), and [this](https://github.com/thoughtbot/carnival/blob/master/config/settings.yml#L19) depending on environment.

jferris: Okay, great. Much better to have Yesod handle this.

PR #142, opened 2015-03-03
Begin allowing Site-specific data

pbrisbin:
Changes:

- Create Site model
- Migrations to create dev/staging/production blog records
- Add `/sites/#SiteId` prefix to most routes
- Render FeedR and EmbedR for a specific Site
- EmbedR uses its SiteId in comments API calls

Deployment:

- Requires migrate-deploy (create-site, insert-thoughtbot-sites)
- Comments will temporarily not work, only until I modify CARNIVAL_HOST on
  staging/production to include the `/sites/#SiteId` prefix
- Zapier hook will need to be changed to include `/sites/#SiteId` prefix

### DIFF
--- a/Handler/Comments.hs
+++ b/Handler/Comments.hs
@@ -38,8 +38,8 @@ instance FromJSON CommentRequest where
 
     parseJSON _ = mzero
 
-postCommentsR :: Handler Value
-postCommentsR = do
+postCommentsR :: SiteId -> Handler Value
+postCommentsR _ = do
     allowCrossOrigin
 
     u <- requireAuth_
@@ -57,8 +57,8 @@ postCommentsR = do
         sendNotification notification
         sendResponseStatus status201 $ object ["comment" .= userComment]
 
-putCommentR :: CommentId -> Handler Value
-putCommentR commentId = do
+putCommentR :: SiteId -> CommentId -> Handler Value
+putCommentR _ commentId = do
     allowCrossOrigin
 
     u <- requireAuth_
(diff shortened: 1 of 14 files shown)


jferris:
This looks like a good approach to me.

Is it difficult to put a rewrite in place for old URLs to make the transition without downtime?

We may need to disable the Zapier hook or something to prevent it from spamming our timeline.

pbrisbin:
> Is it difficult to put a rewrite in place for old URLs to make the transition without downtime?

It's not *too* difficult, but we're talking a few seconds of downtime at most, so I didn't think it was worth it. Do you disagree?

> We may need to disable the Zapier hook or something to prevent it from spamming our timeline

What do you think would be spammed? I thought it just wouldn't work until updated. Are you saying that a feed URL change makes it see all existing entries as new?

jferris:
> What do you think would be spammed? I thought it just wouldn't work until updated. Are you saying that a feed URL change makes it see all existing entries as new?

Yeah, I'm not sure. I've had issues like this before when changing URLs. I *think* it's when the feed entry URLs change, but I'm not sure what it will do if the whole feed seems to change at once.

jferris:
I guess the best thing to do here is probably to just check Buffer after deploying to make sure we didn't queue up hundreds of Tweets or anything.

pbrisbin:
I'm pretty confident this feed is only used for our Slack notifications. We don't Tweet anything based on comments on the blog. I'll try to confirm this.

pbrisbin:
I can't find any zapier integrations...

pbrisbin:
So, there's nothing in Zapier based on this feed. The Slack integration is [here](https://thoughtbot.slack.com/services/2417784262). It looks like updating means removing and re-adding.

My assumption is that when you add an RSS-based integration, any existing entries will be ignored and only new entries will trigger the hook. Considering the worst-case result is a bunch of messages in `#general`, I think I'm OK moving forward with that assumption.

Are you OK with brief downtime during the deploy, or should I implement the temporary redirect before merging?

jferris:
I think this is good to merge. Thanks for following up on my RSS paranoia.

PR #143, opened 2015-03-03
Add commentSite attribute to Comment

pbrisbin:
This is phase 1 of 2. Here, all existing comments are given an incorrect 
SiteId of 1. This is OK because we aren't using it for anything yet. After
deployment, I will manually update all existing comments to have the correct
SiteId. I can then follow up with Phase 2: actually rely on commentSite.

### DIFF
--- a/Handler/Comments.hs
+++ b/Handler/Comments.hs
@@ -11,7 +11,7 @@ import Helper.Auth
 import Helper.Request
 import Helper.Validation
 
-import Control.Monad (mzero)
+import Control.Monad (mzero, void)
 import Text.Markdown
 
 import qualified Data.Text.Lazy as TL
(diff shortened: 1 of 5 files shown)


jferris:
It seems like it wouldn't be crazy to write a migration that would add `SiteId` as `NULL`, set `SiteId` to a valid value, and then change it to be `NOT NULL`.

That seems preferable because it avoids the dummy value and allows you to test out the change locally and then replay it in production.

Why does `1` work as a valid value? Are we not using foreign key constraints?

pbrisbin:
> It seems like it wouldn't be crazy to write a migration that would add SiteId as NULL, set SiteId to valid value, and then change it to be NOT NULL.

I can't picture how this would work. From within a migration, I can't dynamically see what Env I'm in and therefore find the correct Site record to use.

> That seems preferable because it avoids the dummy value and allows you to test out the change locally and then replay it in production.

I can still test things out locally with this approach. In fact, I have.

> Why does 1 work as a valid value? Are we not using foreign key constraints?

1 is a valid value in all environments. It's the "robots-development" Site.

PR #144, opened 2015-03-04
Make all actions site-specific

pbrisbin:
- Fetch comments scoped to the given site
- Use the comment's site to render Email link

Note: aace81a is a big refactor commit to remove all direct uses of the
UserComment constructor, replaced by a smart builder and record accessors. With
that done, eac57a3 (adding a Site field to it) was a much smaller change. I
suggest reviewing those two separately and in isolation.

With this, Carnival fully supports multiple sites run from the same installed
instance.

### DIFF
--- a/Handler/Comments.hs
+++ b/Handler/Comments.hs
@@ -49,10 +49,11 @@ postCommentsR siteId = do
     c <- fmap (buildComment t u siteId) requireJsonBody
 
     runValidation validateComment c $ \v -> do
-        cid <- runDB $ insert v
+        userComment <- runDB $ do
+            commentId <- insert v
+            buildUserComment (Entity commentId v) u
 
-        let userComment = UserComment (Entity cid v) u
-            notification = NewComment userComment
+        let notification = NewComment userComment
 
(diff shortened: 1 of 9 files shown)


jferris:
This looks good to me.

We have a lot of handlers which are `SiteId -> x`, Is there a way to abstract that into a `SiteHandler` type or something?

pbrisbin:
> We have a lot of handlers which are SiteId -> x, Is there a way to abstract that into a SiteHandler type or something?

Sounds like a case for `ReaderT`, except that I don't think we can change the type of Handler functions since they're "auto"-invoked by Yesod as part of routing.

pbrisbin:
Hmm. Thinking about this more, I think you're right. The Handler types wouldn't change, but they'd look something like:

```haskell
postComment :: SiteId -> Handler Value
postComment siteId = withSite siteId $ do
    -- ...
```

Where the second argument to `withSite` would be in a `SiteHandler` and have "global" access to a `Site`. I'm not sure where this would lead, but it could be worth exploring. I think I'll merge this as-is but add an Issue to investigate that -- is should be done in an isolated PR anyway.

PR #146, opened 2015-03-04
Add basic Site administration

pbrisbin:
- Membership model
- CRUD actions on Site
- Users are tied to Sites they create
- Existing User/Site memberships can be created manually

New homepage:

Site listing for create/view:

Edit view:

Clearly, it still needs UI/UX work...


Handler/Sites.hs
```diff
@@ -0,0 +1,77 @@
...
+        _ -> return ()
+
+    sites <- runDB $ findSites userId
+
+    defaultLayout $ do
+        setTitle "Carnival - Sites"
+        $(widgetFile "sites/index")
+
+postSitesR :: Handler Html
+postSitesR = getSitesR
```
jferris: Eh?

pbrisbin: Yeah, I know we discussed this yesterday, but after thinking a bit more I think I'll split the functionality. I'm not sure if it's even an idiom anymore, the examples in the Yesod Book no longer do it this way at least.

pbrisbin:
@jferris I split the POST/GET stuff and renamed that 404-helper.

I'd like to make some more changes after I'm done getting #147 out, then add some warning-ish copy to the home page. Once deployed, anyone who stumbles on this would be able to try it. I want to encourage that, but also be sure they know what they're trying.

If you still want, we can continue going through this together before or after that, just let me know.

Foundation.hs
```diff
@@ -113,6 +116,12 @@ embedLayout widget = do
     pc <- widgetToPageContent widget
     withUrlRenderer $(hamletFile "templates/embed-layout-wrapper.hamlet")
 
+embedExample :: Maybe SiteId -> Widget
```
jferris: Seems a little weird to have this in `Foundation`. Maybe just throw it in its own module?

PR #147, opened 2015-03-05
Add InitR, a single JS script to load Carnival

pbrisbin:
This brings the loading of JS/CSS from Embed into the DOM that was currently
being done directly on the Robots blog into a single initializer.

With this, default usage becomes:

```html
<script src="https://carnival.thoughtbot.com/sites/1/init.js"></script>
<script>
  Carnival.init();
</script>
```
Once deployed, I can make this change to the blog: http://sprunge.us/iVdV?diff


cpytel:
This looks like a good change. :+1:

PR #149, opened 2015-03-05
Add test coverage for Handler.Sites

pbrisbin:


test/Handler/SitesSpec.hs
```diff
@@ -0,0 +1,90 @@
+{-# LANGUAGE OverloadedStrings #-}
+module Handler.SitesSpec where
+
+import TestHelper hiding (createSite)
```
jferris: Should we try to rename this test helpers so that the names won't conflict?

test/Handler/SitesSpec.hs
```diff
@@ -0,0 +1,90 @@
...
+            get SitesR
+
+            htmlNoneContain "li" "Site 1"
+            htmlNoneContain "li" "Site 2"
+
+        it "allows for creation of a site" $ do
+            user <- createUser "1"
+            authenticateAs user
+
+            get SitesR
```
jferris: Is this step necessary? Does state persist between these fake requests?

test/TestHelpers/Request.hs
```diff
@@ -37,3 +38,15 @@ getWithParams url params = request $ do
     setMethod  "GET"
     mapM_ (uncurry addGetParam) params
     setUrl url
+
+postForm :: (Yesod site, RedirectUrl site url)
+         => url
+         -> RequestBuilder site ()
+         -> YesodExample site ()
+postForm url fill = request $ do
```
jferris: This is nice.

test/Handler/SitesSpec.hs
pbrisbin: I'm going to follow up with a more overall change to our helpers and
"factories". I'll definitely remove the hiding as part of that.

test/Handler/SitesSpec.hs
pbrisbin: You need to visit the page containing the CSRF token to be able to post the
form. I could incorporate this into postForm though (especially if I rely
on the GET/POST being the same route).

pbrisbin:
Changed so `postForm` does the GET-then-POST itself.

test/Handler/SitesSpec.hs
pbrisbin: To clarify, I'd like to follow up *after* merging this with another PR.

pbrisbin:
@jferris any further thoughts? I have a new branch cut from this one with general improvements to `test/*`.

jferris:
Looks good to merge!

PR #150, opened 2015-03-10
Better test module organization, etc

pbrisbin:
- Add hspec-expectations-lifted

This removes the need for Helpers.Assertions

- Put generic runDB/authenticateAs helpers into TestImport directly

This (the rename and function move) gets us closer to the default scaffold.

- Add Factories module

This replaces Helpers.DB. All functions live in DB for composability and
functions used by only one Example were moved to be directly with it.

- Rename Helpers.Request to Y.Test.Extension

This will soon become a PR to upstream

- Remove tests for unused-Handlers

They've never been used and only increase maintenance burden. In my next PR,
I'll remove the Handlers themselves (and any code they use).

- Move Helpers.Response directly into CommentsSpec

It is only used there.


test/TestImport.hs
```diff
@@ -0,0 +1,105 @@
+{-# LANGUAGE OverloadedStrings #-}
+{-# LANGUAGE QuasiQuotes #-}
+module TestImport
+    ( withApp
+    , runDB
+    , authenticateAs
+
+    -- Convenience exports
```
pbrisbin: I thought that not using the `import as X`, `export as module X` convention would be better so you can see clearly what's being re-exported. I'm not sure though and may go back to it.

jferris:
> Remove tests for unused-Handlers
> They've never been used and only increase maintenance burden. In my next PR,
> I'll remove the Handlers themselves (and any code they use).

I'd recommend removing those handlers and their tests in a separate pull request on master and then rebasing this on top of the branch for that pull request.

test/Factories.hs
```diff
@@ -0,0 +1,61 @@
...
+        , commentSite = siteId
+        , commentThread = thread
+        , commentArticleTitle = "title"
+        , commentArticleURL = article
+        , commentArticleAuthor = "John Smith"
+        , commentBody = Markdown body
+        , commentCreated = now
+        }
+
+insertEntity :: (PersistEntity e, PersistEntityBackend e ~ SqlBackend)
```
jferris: Weird that there isn't a function like this!

test/Handler/CommentsSpec.hs
```diff
@@ -11,30 +17,32 @@ spec :: Spec
...
         it "returns a list of comments by article" $ do
-            Entity siteId _ <- createSite
-
-            u <- createUser "1"
-            c1 <- createComment (entityKey u) siteId "1" "1" "1"
-            c2 <- createComment (entityKey u) siteId "1" "2" "2"
-            c3 <- createComment (entityKey u) siteId "2" "1" "3"
+            siteId <- runDB $ insert $ buildSite
+            user <- runDB $ createUser "1"
+            (c1, c2, c3) <- runDB $ (,,)
```
jferris: This strikes me as a little too clever. Is the desire to avoid repeating `runDB`? I feel like the repetition is worth avoiding the cognitive overhead of using the tuple/applicative.

test/Handler/CommentsSpec.hs
```diff
@@ -46,17 +54,18 @@ spec = withApp $ do
...
+            siteId <- runDB $ insert $ buildSite
 
             post $ CommentsR siteId
 
             statusIs 401
 
         it "allows commenting by authenticated users" $ do
-            u <- createUser "1"
-            Entity siteId _ <- createSite
+            (siteId, user) <- runDB $ (,)
```
jferris: Look at you, naming your variables! Why did `u` become unacceptable here but not elsewhere?

test/Handler/CommentsSpec.hs
pbrisbin: I personally find this idiom really clear once you get used it. I understand we're not all there yet though, so I'll avoid it for now.

test/Handler/CommentsSpec.hs
pbrisbin: I really can't decide! I'll go through and make things consistent one way or another.

test/Handler/CommentsSpec.hs
pbrisbin: > Is the desire to avoid repeating runDB?

Technically, less `runDB` calls should perform better. Aesthetically, the additional `runDB`s often cause the line to go over 80 and require awkward breaking (this may explain my `user`/`u` waffle as well). I'll still remove the Applicative, but that was the true reasoning.

test/Handler/SitesSpec.hs
```diff
@@ -74,15 +81,17 @@ spec = withApp $ do
             statusIs 303 -- redirect
 
             Just site' <- runDB $ DB.get $ entityKey site
-            assertEqual' "new-name" $ siteName site'
```
jferris: Ha, this is nice if only because we don't have `assertEqual'` in there anymore.

test/TestImport.hs
jferris: Is the advantage to `as X` that you don't need to re-import it below?

pbrisbin: No, it's this:

```haskell
module Foo
    ( foo
    , bar
    , module X
    )

import Bar as X
import Baz as X
import Bat as X
```

Vs:

```hs
module Foo
    ( foo
    , bar
    , module Bar
    , module Baz
    , module Bat
    )

import Bar
import Baz
import Bat
```

Benefit would be less churn when changing what you re-export, downside would be you need to look at the `import`s to see what that `X` means. Keeping them grouped can help, so it's might be fine.

test/Handler/CommentsSpec.hs
jferris: Is there a reason we could use these without `runDB` and we can't now?

pbrisbin:
> I'd recommend removing those handlers and their tests in a separate pull request on master

FYI: #151

test/Factories.hs
pbrisbin: https://github.com/yesodweb/persistent/issues/376

test/Handler/CommentsSpec.hs
pbrisbin: Yes, their types changed from `YesodExample App a` to `DB a`. They previously called `runDB` internally and no longer do. This has a few benefits:

- `runDB` can live separately from `Factories` without a cyclic import
- multiple factory functions can be composed into a single `runDB` call (there are cases where the Applicative cleverness is not needed to achieve this)

Eventually, I want to take this further and have clear separation between `runDB`, `insert` and `buildFoo` within any `createFoo` functions. This is the only way we can get rid of the multiple arguments to `createComment` which I've always found obfuscating.

For example,

```hs
-- we often only care about Thread, but we have to pass everything always
createComment "1" "1" "1"

-- I'd like this kind of composability instead
runDB $ insert $ buildComment { commentThread "1" }
```

`createComment` could still exist as `runDB $ insert buildComment` for cases where you don't care about the comment data itself.

jferris: Cool. I think it will be good to play around until we get a style we're happy with here. The stuff about `runDB`/`YesodExample` is good to know.

pbrisbin:
Build is failing due to this: https://github.com/mietek/halcyon/issues/36

mietek:
On it.

mietek:
https://github.com/mietek/halcyon/issues/36 should now be fixed.  The fix is now tested and merged to `master`.

Please try again, keeping in mind if you’ve disabled self-updates, you’ll need to re-enable them, or update Halcyon manually.

pbrisbin:
@jferris rebased, green build (before rebase), ready for a final look.

pbrisbin:
Oh, thanks thanks for the quick fix @mietek !

jferris:
Looks good to merge.

PR #152, opened 2015-03-11
Add a demo route

pbrisbin:
This removes the need to run the blog locally and provides a second Site on
which to test Carnival.

The commenting does load and work, but the positioning is incorrect and the
indicators appear off-screen and cannot be interacted with. I'll follow up
separately with a PR/PRs to address these issues and, once done, add a link to
DemoR from the homepage.

### DIFF
--- a/Handler/Demo.hs
+++ b/Handler/Demo.hs
@@ -2,11 +2,9 @@ module Handler.Demo where
 
 import Import
 
-import Yesod.Default.Config
-
 getDemoR :: Handler Html
 getDemoR = do
-    root <- fmap (appRoot . settings) getYesod
+    root <- getAppRoot
     Entity siteId _ <- runDB $ upsert (demoSite root) []
 
     defaultLayout $ do
--- a/Handler/Demo.hs
+++ b/Handler/Demo.hs
@@ -1,0 +1,24 @@
+module Handler.Demo where
+
+import Import
+
+import Yesod.Default.Config
+
+getDemoR :: Handler Html
+getDemoR = do
+    root <- fmap (appRoot . settings) getYesod
+    Entity siteId _ <- runDB $ upsert (demoSite root) []
+
+    defaultLayout $ do
+        setTitle "Carnival Demo"
+        $(widgetFile "demo")
+
+demoSite :: Text -> Site
+demoSite root = Site
+    { siteName = "carnival-demo"
+    , siteBaseUrl = root
+    , siteRssAuthor = "Carnival Demo"
+    , siteRssTitle = "Comments on Carnival Demo"
+    , siteRssDescription = "Comments on Carnival Demo"
+    , siteRssLanguage = "en-us"
+    }
--- a/Handler/Sites.hs
+++ b/Handler/Sites.hs
@@ -86,7 +86,7 @@ embedExample msiteId = do
 siteForm :: Maybe Site -> Form Site
 siteForm msite = renderBootstrap3 BootstrapBasicForm $ Site
     <$> areq textField (bfl "Name") (siteName <$> msite)
-    <*> areq textField (bfl "Base URL") (siteBaseUrl <$> msite)
+    <*> areq baseUrlField (bfl "Base URL") (siteBaseUrl <$> msite)
     <*> areq textField (bfl "RSS Author") (siteRssAuthor <$> msite)
     <*> areq textField (bfl "RSS Title") (siteRssTitle <$> msite)
     <*> areq htmlField (bfl "RSS Description") (siteRssDescription <$> msite)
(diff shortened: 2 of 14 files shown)


Handler/Demo.hs
```diff
@@ -0,0 +1,35 @@
...
+    { siteName = "carnival-demo"
+    , siteBaseUrl = root
+    , siteRssAuthor = "Carnival Demo"
+    , siteRssTitle = "Comments on Carnival Demo"
+    , siteRssDescription = "Comments on Carnival Demo"
+    , siteRssLanguage = "en-us"
+    }
+
+getAdmin :: DB UserId
+getAdmin = entityKey . fromJust <$> getBy (UniqueUser "github" "50812")
```
jferris: Will this always exist, or do we need to create it as above?

Seems kind of strange to embed this ID in here.

Handler/Demo.hs
```diff
@@ -0,0 +1,35 @@
...
+    siteId <- runDB $ findDemoSite root
+
+    defaultLayout $ do
+        setTitle "Carnival Demo"
+        $(widgetFile "demo")
+
+findDemoSite :: Text -> DB SiteId
+findDemoSite root = do
+    admin <- getAdmin
+    msite <- listToMaybe <$> findSites admin [SiteBaseUrl ==. root]
```
jferris: Is there a `findOrInsert` helper we could extract here?

Handler/Demo.hs
pbrisbin: This is me and I exist everywhere... except other developer's machines. I had forgotten about that part.

It is strange, but I can't think of how else to do this. Should I create a Nobody user to own the Site? Also (in addition to that question), would it be better to do this in a migration?

Handler/Demo.hs
pbrisbin: I chose to re-use `findSites` vs extracting something more general that wouldn't be reused anyway. There is a `findOrInsert` in persistent already I think, but we couldn't use it because of the two tables involved (`site`, and `membership`). I can't picture a `findOrInsert` of much generic value that correctly handles the "join" lookup.

Handler/Demo.hs
pbrisbin: I know, I'll just create a Site with no members.

pbrisbin:
Ready for a second look

Handler/Demo.hs
```diff
@@ -0,0 +1,37 @@
...
+        $(widgetFile "demo")
+
+findDemoSite :: Text -> DB SiteId
+findDemoSite root = do
+    sites <- selectList [SiteBaseUrl ==. root] []
+    msite <- listToMaybe <$> filterM noMembers sites
+
+    maybe (createDemoSite root) (return . entityKey) msite
+
+  where
```
jferris: Is there a reason we need to make sure there aren't any members? Won't any site that points to the root URL be okay?

jferris:
I had one follow-up question.

Handler/Demo.hs
pbrisbin: Theoretically, then anyone could create a site with the root URL and it might become the site used by `/demo`. That just sounds dangerous in terms of abuse. I don't think there's anything a user could do now but who knows how `Site` evolves and what control of the demo site could mean in the future; enforcing that no one can administer this sites feels safer.

I'm probably being paranoid though. If you think it'd be fine, I can take it out. I assume you do?

jferris: Yeah, I think it's fine to take it out.

Do we enforce uniqueness on site URL? We probably should. If we do, it won't be possible for somebody to override the site once we've created it.

pbrisbin: Yeah, I'll add the uniqueness constraint. I had been considering it but hadn't done it yet. Given that, I'll feel better about not enforcing no-members here.

pbrisbin:
OK all set. I'm using `upsert` without updates as a `findOrCreate`.

jferris:
Looks good.

Handler/Sites.hs
```diff
@@ -97,6 +96,14 @@ siteForm msite = renderBootstrap3 BootstrapBasicForm $ Site
     bfl :: Text -> FieldSettings site
     bfl = bfs
 
+    baseUrlField :: Field Handler Text
+    baseUrlField = flip checkM textField $ \url -> do
```
pbrisbin: I realized the uniqueness constraint would result in a poor UX now because, by default, the PG_ERROR just comes through as a 500. I decided to add this validation even though it's a race condition. It improves things a lot with little effort. I may or may not add an Issue to look into doing it without the race condition (maybe rescuing the IOException and manually adding an error to the form before re-rendering).

jferris: For what it's worth, this is what we do on Rails app (allow a race condition).

No valid data is allowed into the database, and the only way somebody will trigger the race condition is by rapidly resubmitting the form.

PR #160, opened 2015-03-11
Feed-related improvements

pbrisbin:
- Add placeholders to better explain create-site fields
- Add link to Feed in sites/show

### DIFF
--- a/Handler/Feed.hs
+++ b/Handler/Feed.hs
@@ -24,10 +24,10 @@ feedFromComments (Entity siteId site) comments = do
     render <- getUrlRender
 
     rssFeedText Feed
-        { feedAuthor      = siteRssAuthor site
-        , feedTitle       = siteRssTitle site
-        , feedDescription = siteRssDescription site
-        , feedLanguage    = siteRssLanguage site
+        { feedAuthor      = siteName site
+        , feedTitle       = "Comments on " <> siteName site
+        , feedDescription = toHtml $ "Comments on " <> siteName site
+        , feedLanguage    = siteLanguage site
         , feedLinkSelf    = render $ FeedR siteId
         , feedLinkHome    = siteBaseUrl site
         , feedUpdated     = getCommentCreated $ head comments
--- a/Handler/Sites.hs
+++ b/Handler/Sites.hs
@@ -72,7 +72,7 @@ getDeleteSiteR siteId = do
 postDeleteSiteR :: SiteId -> Handler ()
 postDeleteSiteR siteId = do
     void $ requireMemberSite siteId
-    runDB $ destroySite $ siteId
+    runDB $ destroySite siteId
     setMessage "Site deleted!"
     redirect SitesR
 
@@ -91,7 +91,7 @@ siteForm msite = renderBootstrap3 BootstrapBasicForm $ Site
   where
     fs :: Text -> Text -> FieldSettings site
     fs msg placeholder = FieldSettings
-        { fsLabel = (SomeMessage msg)
+        { fsLabel = SomeMessage msg
         , fsTooltip = Nothing
         , fsId = Nothing
         , fsName = Nothing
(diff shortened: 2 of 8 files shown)


Handler/Sites.hs
```diff
@@ -85,17 +85,25 @@ embedExample msiteId = do
...
-    <$> areq textField (bfl "Name") (siteName <$> msite)
-    <*> areq textField (bfl "Base URL") (siteBaseUrl <$> msite)
-    <*> areq textField (bfl "RSS Author") (siteRssAuthor <$> msite)
-    <*> areq textField (bfl "RSS Title") (siteRssTitle <$> msite)
-    <*> areq htmlField (bfl "RSS Description") (siteRssDescription <$> msite)
-    <*> areq textField (bfl "RSS Language") (siteRssLanguage <$> msite)
+    <$> areq textField (fs "Name" "my-site") (siteName <$> msite)
+    <*> areq textField (fs "Base URL" "https://example.com") (siteBaseUrl <$> msite)
+    <*> areq textField (fs "Comments Feed Author" "My Site") (siteRssAuthor <$> msite)
+    <*> areq textField (fs "Comments Feed Title" "My Site Comments") (siteRssTitle <$> msite)
```
jferris: It still seems like it would be an improvement to standardize/generate these two fields. I'm not sure why somebody would want to customize the description for their comments feed.

pbrisbin: That's fine, it just means I can't do the form in this simple way. I'll have to accept a datatype with fewer fields, then build a full Site out of the result.

Which fields do you think they should enter: just Name, URL, Language?

I'm going to have to rename our sites I think ("Latest comments on robots-production" doesn't sound right).

jferris: > That's fine, it just means I can't do the form in this simple way. I'll have to accept a datatype with fewer fields, then build a full Site out of the result.

I was thinking we'd remove those from the `Site` type entirely and then just generate the description, etc in the feed handler.

> Which fields do you think they should enter: just Name, URL, Language?

Yeah, I think that will be fine. We could probably reuse the site title as the author name for MVP.

> I'm going to have to rename our sites I think ("Latest comments on robots-production" doesn't sound right).

Does anybody look at that but us? I can't think of how somebody would use this except to notify themselves of new comments through something like Zapier, so I don't think the values are very important.

pbrisbin: Ah OK. Yeah, that all makes sense.

pbrisbin:
Ready for re-review.

Handler/Feed.hs
```diff
@@ -24,10 +24,10 @@ feedFromComments (Entity siteId site) comments = do
     render <- getUrlRender
 
     rssFeedText Feed
-        { feedAuthor      = siteRssAuthor site
-        , feedTitle       = siteRssTitle site
-        , feedDescription = siteRssDescription site
-        , feedLanguage    = siteRssLanguage site
+        { feedAuthor      = siteName site
+        , feedTitle       = "Comments on " <> siteName site
+        , feedDescription = toHtml $ "Comments on " <> siteName site
```
jferris: Do we need to HTML escape `siteName` here?

jferris:
Looks good.

Handler/Feed.hs
pbrisbin: I'm sure `toHtml` is going to do the Right Thing.

jferris: Sure, I just didn't know if `toHtml` meant "take this literal string and mark it as HTML" or "escape HTML characters in this literal string."

pbrisbin: I'll double-check, I'm pretty sure you have to use a `prescapedToHtml` or something to avoid escaping.

pbrisbin:
```
λ> toHtml "Foo & Bar"
Foo &amp; Bar
```

I used the `String` instance in this test, because the `Text` one wasn't in scope. I'm sure they behave the same w.r.t. escaping.

Issue #154, opened 2015-03-11
Dummy login doesn't work in development


pbrisbin:
The Dummy login builds a `Creds` with the "dummy" `credsPlugin` and no `credsExtra`. This doesn't work in development because we require a name/email (provided by GitHub) to be present in the `credsExtra`. This *does* work in test because we create a user ahead of time. Existing users aren't found in development because they have the "github" plugin. So the auth flow tries to create them new with the "dummy" plugin and fails because name/email is not present.

We should detect a "dummy" login within `getAuthId` and do something to make it work again.

Optionally, we should add some logging around this area is it's been difficult to debug why a login doesn't work.

pbrisbin:
Fixed as part of #161

PR #161, opened 2015-03-12
Update to latest scaffold

pbrisbin:
Benefits:

- ClassyPrelude
- Better, ENV-based configuration
- Better development tools (handler/db)
- Better logging format
- Consist, stylish-haskell-formatted source
- Remove some unused files

Notes:

The process went like this:

- Clone the latest scaffold and checkout the postgres branch
- Update our cabal file to match the scaffold
- Remove all halcyon constraint files
- Remove the --enable-test compilation flag (this stops halcyon from building
  carnival itself as part of the sandbox step)
- bin/halcyon-build --dependencies-only
- cp missing files from yesod-scaffold
- rm extra files from current directory
- vimdiff each shared file; adjust as needed
- Try to compile Application.hs; fix errors
- Add --enable-test back, bin/halcyon-build, bring back constraints files
- Try yesod devel / yesod test; fix errors

I reformatted some code (Application.hs, TestImport.hs) to match the scaffold
even though it was already functionally the same. This made this update and
should make future updates easier.

I made it so that lists (exports, imports, dependencies, extensions, etc) match
the scaffold exactly, then a blank line, then anything additional we need
(alphabetized). There were only a few places where it didn't make sense to
follow this exactly (e.g. TestImport's import Test.Hspec), but I did attempt to
keep a vimdiff against the scaffold as readable as possible. This and the
scaffold's use of stylish-haskell means that this code-base doesn't match our
style guide completely -- I think that's OK.


.ghci
```diff
@@ -1,7 +1,21 @@
...
+:set -XGeneralizedNewtypeDeriving
+:set -XMultiParamTypeClasses
+:set -XNoImplicitPrelude
+:set -XNoMonomorphismRestriction
+:set -XOverloadedStrings
+:set -XQuasiQuotes
+:set -XRecordWildCards
+:set -XTemplateHaskell
+:set -XTupleSections
+:set -XTypeFamilies
```
pbrisbin: The scaffold puts all extensions on one line and is currently missing NoImplicitPrelude. I'm going to PR the repo to add that and use this multi-line format which I think is easier to manage.

jferris: Good idea.

Application.hs
```diff
@@ -47,92 +50,160 @@ import Handler.Demo
...
 -- migrations handled by Yesod.
-makeApplication :: AppConfig DefaultEnv Extra -> IO Application
-makeApplication conf = do
-    foundation <- makeFoundation conf
+makeFoundation :: AppSettings -> IO App
+makeFoundation appSettings = do
+    loadEnv
+
+    dbconf <- if appDatabaseUrl appSettings
+        then postgresConf 10
```
pbrisbin: TODO: use the poolsize from appDatabaseConf, then we can set `PGPOOLSIZE=100` on staging/production. We've been using a hardcode `10` currently anyway, but that'll be a nice fix.

Handler/Feed.hs
```diff
@@ -30,7 +30,7 @@ feedFromComments (Entity siteId site) comments = do
         , feedLanguage    = siteLanguage site
         , feedLinkSelf    = render $ FeedR siteId
         , feedLinkHome    = siteBaseUrl site
-        , feedUpdated     = getCommentCreated $ head comments
+        , feedUpdated     = getCommentCreated $ P.head comments
```
jferris: Does `unsafeHead` work here?

templates/login.hamlet
```diff
@@ -1,12 +1,3 @@
-$if development
```
jferris: It's not clear to me why this was removed.

jferris:
Nice, it will be good to work with the latest Yesod stuff, especially classy-prelude. The settings changes are interesting, too.

Settings.hs
```diff
@@ -32,28 +33,81 @@ type PersistConf = PostgresConf
...
+    -- ^ Google Analytics code
+    , appSendMail               :: Bool
+    -- ^ Actually send e-mail (via SendGrid)?
+    , appDatabaseUrl            :: Bool
+    -- ^ Parse database info from DATABASE_URL?
+    , appAllowDummyAuth         :: Bool
+    -- ^ Add the Dummy Auth plugin (for authenticating in tests)?
+    }
+
+instance FromJSON AppSettings where
```
jferris: Interesting that `FromJSON` is used even though the settings are actually YAML.

templates/login.hamlet
pbrisbin: Oh, I meant to mention that... Since the new scaffold doesn't have `development`, I had to do something here. I had plans to remove this for reasons I'll go into, so I just did it now.

Basically, the Dummy Login is useful only in test. It allows us to authenticate as a given user in a test by POSTing directly to the dummy auth route (`TestImport.authenticateAs`). It does not go by way of this form, so removing it doesn't impact that.

In development, it doesn't work anyway for the reasons stated here in #154.

Even if I fixed that issue, this would still not be useful because the user must already exist for the Dummy login to work. For the user to already exist, you must've set up GitHub auth. If you've set up GitHub auth, you might as well use the actual GitHub login.

The only place the Dummy login would be useful is if you were trying to develop without an internet connection. I don't see that as a reason to keep it.

Handler/Feed.hs
pbrisbin: I'd imagine so, that's what I was going to look for.

templates/login.hamlet
jferris: Okay, good to know. I agree with removing it.

Settings.hs
pbrisbin: Yeah, it's a bit surprising at first but nice once you're used to it. Rather than writing a `FromYAML` thing, the library generically parses Yaml->JSON then as an end-user it's just like using aeson.

pbrisbin:
@jferris any more thoughts before I merge this and try it on staging?

jferris:
No, this looks good to me.

mietek:
FYI — you should be able to use [`HALCYON_IGNORE_ALL_CONSTRAINTS`](https://halcyon.sh/reference/#halcyon_ignore_all_constraints) instead of having to remove constraints files.

PR #163, opened 2015-03-12
Get Demo page working

pbrisbin:
These changes get comments actually working on the Demo page. Along with
https://github.com/thoughtbot/maitre-d/pull/103, the blog is unaffected.

### DIFF
--- a/templates/demo.hamlet
+++ b/templates/demo.hamlet
@@ -2,19 +2,82 @@
 
 <article>
   <p>
+
     This is a demo of an article with Carnival loaded. Each paragraph can be
-    commented on individually.
+    commented on individually. It uses no additional styling and has the minimal
+    snippet of JavaScript needed to install Carnival.
 
   <p>
-    This is a second paragraph. Try adding comments to both.
+
+    Lorem ipsum dolor sit amet, consectetur adipiscing elit, sed do eiusmod
+    tempor incididunt ut labore et dolore magna aliqua. Ut enim ad minim veniam,
+    quis nostrud exercitation ullamco laboris nisi ut aliquip ex ea commodo
+    consequat. Duis aute irure dolor in reprehenderit in voluptate velit esse
+    cillum dolore eu fugiat nulla pariatur. Excepteur sint occaecat cupidatat
+    non proident, sunt in culpa qui officia deserunt mollit anim id est laborum.
 
   <p>
-    View the page source to see how simple it was to install.
 
-  <pre>
-    \ def message
-    \   "You can even comment on code"
-    \ end
+    Sed ut perspiciatis unde omnis iste natus error sit voluptatem accusantium
+    doloremque laudantium, totam rem aperiam, eaque ipsa quae ab illo inventore
+    veritatis et quasi architecto beatae vitae dicta sunt explicabo. Nemo enim
+    ipsam voluptatem quia voluptas sit aspernatur aut odit aut fugit, sed quia
+    consequuntur magni dolores eos qui ratione voluptatem sequi nesciunt. Neque
+    porro quisquam est, qui dolorem ipsum quia dolor sit amet, consectetur,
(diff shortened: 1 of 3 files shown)


templates/embed.lucius
```diff
@@ -7,8 +7,9 @@
 
 .carnival-comment-indicator {
   cursor: pointer;
-  position: absolute;
-  right: -30px;
+  float: right;
+  position: relative;
+  right: -15px;
```
pbrisbin: This is the only way I could find to do this without relying on the specifics of exactly where and how wide the user's `article` element is. It does cause some minor text-re-flow when the indicators get added to the DOM though. Unfortunately, I don't see a way to address this without re-working the entire approach (which I think we should do eventually).

templates/demo.hamlet
```diff
@@ -2,19 +2,82 @@
...
+    dolor repellendus. Temporibus autem quibusdam et aut officiis debitis aut
+    rerum necessitatibus saepe eveniet ut et voluptates repudiandae sint et
+    molestiae non recusandae. Itaque earum rerum hic tenetur a sapiente
+    delectus, ut aut reiciendis voluptatibus maiores alias consequatur aut
+    perferendis doloribus asperiores repellat.
 
   <p>
-    View the page source to see how simple it was to install.
 
-  <pre>
```
jferris: Do we not want a code example?

templates/embed/Article.coffee
```diff
@@ -39,3 +41,18 @@ class Article
...
+      @element.style.position = 'relative'
+      @element.style.left = '-150px'
+
+  restoreArticle: =>
+    if @wideScreen()
+      @element.style.position = @prevPosition
+      @element.style.left = @prevLeft
+
+  wideScreen: =>
+    not window.matchMedia('only screen and (max-width: 640px)').matches
```
pbrisbin: http://caniuse.com/#feat=matchmedia

templates/demo.hamlet
pbrisbin: This whole thing'll go away when we find a real page to add comments to.

pbrisbin:
With the article `position: relative`, I was able to get things working with only the two minor changes that we were doing manually over on the blog anyway. This is good because it means I don't have to do a synchronized deploy.

Removing the `float: right` on the form was a preference thing, I think it looks better without it.

templates/embed/Article.coffee
```diff
@@ -39,3 +42,15 @@ class Article
     container.className = 'carnival'
     @element.insertBefore(container, @element.firstChild)
     container
+
+  shiftArticle: =>
+    if @wideScreen()
+      @prevLeft = @element.style.left
+      @element.style.left = '-150px'
```
jferris: Should we subtract 150 from the previous left?

pbrisbin: Sure. I wonder if `@element.style.left` could return `"200em"`...

jferris: This is the Internet! Anything can happen!

pbrisbin:
@jferris is this good to go?

test/Handler/CommentsSpec.hs
```diff
@@ -43,7 +42,7 @@ spec = withApp $ do
 
             get $ CommentsR siteId
 
-            valueEquals =<< commentsResponse siteId u [c2, c3, c1]
+            valueEquals =<< commentsResponse u [c2, c3, c1]
```
pbrisbin: I must've botched this when I rebased the feed-fix over the comment-order fix.

test/Handler/CommentsSpec.hs
```diff
@@ -6,7 +6,6 @@ module Handler.CommentsSpec
 import TestImport
 import Model.UserComment
 import Data.Aeson (Value, (.=), encode, object)
-import Data.Time (getCurrentTime)
```
pbrisbin: ClassyPrelude seems to have a lot more out of the box than Prelude.

PR #164, opened 2015-03-12
findUserComments should sort by CreatedAt

pbrisbin:
Whoops...

### DIFF
--- a/Model/UserComment.hs
+++ b/Model/UserComment.hs
@@ -49,7 +49,7 @@ findUserComments siteId marticle = do
             , fmap (CommentArticleURL ==.) marticle
             ]
 
-    selectWithUsers filters []
+    selectWithUsers filters [Asc CommentCreated]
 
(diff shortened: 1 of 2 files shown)


jferris:
But it compiled!!!

Should we have any tests for this change?

pbrisbin:
This is very strange. In trying to add a test, I'm finding that I can't make it fail.

```
runDB $ selectList [] []
```

Seems to naturally sort by Created.

pbrisbin:
And yet, this is returning comments out of order: https://carnival-staging.herokuapp.com/sites/2/comments?article=announcing-maintaining-open-source-projects-book

pbrisbin:
I'm done figuring this out, it has to be something about my Postgres setup. I'm just going to add the test even though I can't make it fail.

pbrisbin:
@jferris is this OK to merge?

PR #169, opened 2015-03-13
Only shift Article if required

pbrisbin:
When moving from paragraph to paragraph with commenting open, we were shifting
the article further and further, this guards the shifting to only occur if
needed.


jferris:
I'm unfamiliar with `classList`, but this makes sense to me.

pbrisbin:
As for compatibility it looks to be the same as matchMedia (all but IE9).

PR #172, opened 2015-03-16
Remove custom error handling in getOAuthKeys

pbrisbin:
Some drive-by refactors I made

### DIFF
--- a/Application.hs
+++ b/Application.hs
@@ -29,7 +29,7 @@ import System.Log.FastLogger                (defaultBufSize, newStdoutLoggerSet,
                                              toLogStr)
 
 import LoadEnv
-import System.Environment (lookupEnv)
+import System.Environment (getEnv)
 import Web.Heroku.Persist.Postgresql (postgresConf)
 
 -- Import all relevant handler modules here.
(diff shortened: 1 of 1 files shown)


Application.hs
```diff
@@ -93,22 +93,12 @@ makeFoundation appSettings = do
...
-                    { oauthKeysClientId     = pack clientId
-                    , oauthKeysClientSecret = pack clientSecret
-                    }
-
-            _ -> error
-                    $  plugin
-                    <> "_OAUTH_CLIENT_ID or "
-                    <> plugin
-                    <> "_OAUTH_CLIENT_SECRET not set"
+    getOAuthKeys plugin = OAuthKeys
```
gabebw: What is "plugin"? What's an example value? Is there a better name for it?

pbrisbin: > What is "plugin"?

A Yesod `AuthPlugin` name. In this case it's assumed to be pre-capitalized and used as a prefix for environment variables.

> What's an example value

At the call site, `"GITHUB"` is used.

> Is there a better name for it?

It is and has been fine with me so far. Do you have any suggestions now that you know what it is?

gabebw: Is `pluginName` better, or am I stuck in Ruby naming and I should get hip to Haskell's single-variable names again?

pbrisbin: > I stuck in Ruby naming and I should get hip to Haskell's single-variable names again?

This is absolutely how I feel, I'd prefer to name it `p` or `n`, myself :)

gabebw: Ha, this is fine, then.

Foundation.hs
```diff
@@ -130,14 +130,16 @@ instance YesodAuth App where
     -- Override the above two destinations when a Referer: header is present
     redirectToReferer _ = True
 
-    getAuthId creds = runDB $ do
-        muser <- getBy $ UniqueUser (credsPlugin creds) (credsIdent creds)
+    getAuthId Creds{..} = runDB $ do
```
pbrisbin: I'm going to extract this into its own module and test it before making changes. This is a pretty critical path.

gabebw: Yeah, why remove `buildUser`? I liked having it as small methods.

pbrisbin: In this case, `buildUser` was dangerous because `Foundation` doesn't limit it's exports (bit of a long story, but life is far more convenient that way). It's too generic a name to be global, IMO.

pbrisbin: https://github.com/thoughtbot/carnival/pull/188

PR #176, opened 2015-03-16
First draft of /docs

pbrisbin:
- Move DemoR to DocsR
- Add highlight.js
- Write /docs content (suggestions welcome!)
- Add link to navigation bar

PR #178, opened 2015-03-16
WIP: Ensure Settings.StaticFiles is always compiled

pbrisbin:


pbrisbin:
Pulled over from #176 -- see there for the problem. This branch should pass because nothing was added to /static here, but this did not fix this issue.

pbrisbin:
Spoke with Halcyon maintainer -- the reason this didn't work is that the source-dir (`$2`) is not the correct directory within which to `touch`, I should use build-dir (`$3`). This should work in that `Settings.StaticFiles` will be re-compiled on every build, but it's a bit over-aggressive. We should only need to do it when `static/*` is changed. He's offered to add a new hook that will allow us to better decided if we need to perform the `touch`.

pbrisbin:
Going to commit that change directly in pb-docs, as we won't know that it worked without an actual source change to `static/*`.

PR #182, opened 2015-03-18
Don't validate URL uniqueness if unchanged

### LINKED ISSUE #181: Updates fail because of unique validation
Need to disable the unique-base-url validation on PUT requests

pbrisbin:
On updates that don't change URL, the uniqueness validation was causing an
error because the record itself was showing up as an existing record.

This patch checks first if the URL has been changed (or is new) before 
validating uniqueness.

Resolves #181

### DIFF
--- a/Handler/Sites.hs
+++ b/Handler/Sites.hs
@@ -102,11 +102,19 @@ siteForm msite = renderBootstrap3 BootstrapBasicForm $ Site
 
     baseUrlField :: Field Handler Text
     baseUrlField = flip checkM textField $ \url -> do
-        n <- runDB $ count [SiteBaseUrl ==. url]
+        if unchanged url
+            then return $ Right url
+            else ensureUnique url
 
-        return $ case n of
-            0 -> Right url
-            _ -> Left ("This URL is already in use" :: Text)
+      where
+        unchanged url = fmap siteBaseUrl msite == Just url
+
+        ensureUnique url = do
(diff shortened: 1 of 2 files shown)


Handler/Sites.hs
```diff
@@ -102,11 +102,19 @@ siteForm msite = renderBootstrap3 BootstrapBasicForm $ Site
...
-        n <- runDB $ count [SiteBaseUrl ==. url]
+        if unchanged url
+            then return $ Right url
+            else ensureUnique url
 
-        return $ case n of
-            0 -> Right url
-            _ -> Left ("This URL is already in use" :: Text)
+      where
+        unchanged url = fmap siteBaseUrl msite == Just url
```
JoelQ: What is `msite` and where does it come from?

pbrisbin: Functions in a `where` clause have access to the enclosing function's arguments. This is the `msite` argument passed to `siteForm`.

pbrisbin: I usually avoid this unless since, as you noticed, it can be unclear. I considered it worth it in this case vs naming and passing an explicit argument (e.g. `msite'` or something).

JoelQ:
:+1:

pbrisbin:
Thanks @JoelQ -- Travis has been failing lately, unable to get the postgresql connection. I've restarted the build and will rebase/merge when green.

PR #183, opened 2015-03-18
Don't respond 404 for a feed of no comments

### LINKED ISSUE #180: FeedR gives 404 when there are no comments
Need should render an empty feed instead. Currently getting feedUpdated from the first comment, why not just use "now"?

pbrisbin:
If empty:

- Use the current time as updated
- Render a feed with no items

Resolves #180


JoelQ:
This looks right (although my haskell experience is pretty limited)

PR #185, opened 2015-03-18
Enable analytics via Segment

### LINKED ISSUE #174: Setup Google Analytics
Just adding/un-commenting our code in settings.yml should do it. Actually setting up the site in GA would be the real work here.

pbrisbin:
Resolves #174

### DIFF
--- a/config/settings.yml
+++ b/config/settings.yml
@@ -30,4 +30,4 @@ database:
   poolsize: "_env:PGPOOLSIZE:10"
 
 copyright: "thoughtbot, inc. 2015"
-#analytics: UA-YOURCODE
+analytics: "nMWJWvFASCtQm0dr3nIZLDszgHvm7vhx"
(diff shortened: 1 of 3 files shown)


tonyd256:
:+1: sweet JS bro

config/settings.yml
```diff
@@ -30,4 +30,4 @@ database:
   poolsize: "_env:PGPOOLSIZE:10"
 
 copyright: "thoughtbot, inc. 2015"
-#analytics: UA-YOURCODE
+analytics: "nMWJWvFASCtQm0dr3nIZLDszgHvm7vhx"
```
JoelQ: Is this sensitive information?

pbrisbin: This is a write-only code. It's also going into the page source (by design), so I don't see putting it in the repo as a problem.

Please correct me if I'm wrong.

PR #186, opened 2015-03-18
Make siteLanguage a select

### LINKED ISSUE #175: siteLanguage input should use a select
Acceptable values: http://www.rssboard.org/rss-language-codes

pbrisbin:
Resolves #175


Handler/Sites.hs
```diff
@@ -85,7 +86,7 @@ siteForm :: Maybe Site -> Form Site
 siteForm msite = renderBootstrap3 BootstrapBasicForm $ Site
     <$> areq textField (fs "Name" "my-site") (siteName <$> msite)
     <*> areq baseUrlField (fs "Base URL" "https://example.com") (siteBaseUrl <$> msite)
-    <*> areq textField (fs "Language" "en-us") (siteLanguage <$> msite)
+    <*> areq languageField (fs "Language" "") (Just $ fromMaybe "en-us" $ siteLanguage <$> msite)
```
jferris: Not sure if it's better, but I think you can use `Alternative` to do this:

``` haskell
siteLanguage <$> msite <|> "en-us"
```

pbrisbin: That is better (but needs to be `<|> Just "en-us"`

test/Handler/SitesSpec.hs
```diff
@@ -108,3 +111,9 @@ spec = withApp $ do
 
 createSite_ :: UserId -> Site -> DB ()
 createSite_ userId = void . createSite userId
+
+selectLanguage :: Text -> RequestBuilder m ()
+selectLanguage lang = do
```
jferris: Is there no general helper to select an option by label or value? We have to select by index?

pbrisbin: > Is there no general helper to select an option by label or value?

Right. This would be great to add, but I don't know if its existence yet.

> We have to select by index?

Or we could write something that does CSS queries to find the `select` by label (after reproducing Yesod.Test's internal `nameFromLabel` function) then find within it an the correct `option`, then using that `option`'s value. This just seemed easier since `langList` is already available and Yesod creates indexed option-values.

jferris: Cool. If we find that helper later, we can use it. If not, this could be a good pull request to Yesod.Test down the line.

pbrisbin: > this could be a good pull request to Yesod.Test down the line

Definitely. This Friday, I'm going to work on https://github.com/yesodweb/yesod/issues/956, but moving this (and everything else in our Yesod.Test.Extension) to a PR could be good to do next Friday.

pbrisbin:
Renamed `langList` to `languageList` for consistency with `languageField` and rebased. Is this good to go @jferris ?

jferris:
This looks good to me.

PR #188, opened 2015-03-18
Extract and test authenticateUser

pbrisbin:
This code has actually been a source of bugs. With the new scaffold's `Import.NoFoundation` we're able to extract more code out of `Foundation` without running into circular imports. Moving this to `Model.User` made it easy to test.


pbrisbin:
Hmm, looks like this is changing behavior that isn't caught but the test coverage I added. Will have to dig in some more before this is ready for review

pbrisbin:
So the difference is the known user / missing profile case. If we find the user, but the profile's empty we still consider that valid athentication. Will add a test case and fix.

PR #189, opened 2015-03-18
Remove requireAuth_ and Helper.Auth

pbrisbin:
- New yesod-auth responds correctly to JSON requires
- Requires we use an Accept header in specs that see this behavior
- Move requireMemberSite to the only place it's needed

This gets rid of a poorly-named "Helper" module and doesn't recreate a feature
already present in the framework.


Helper/Auth.hs
```diff
@@ -1,29 +0,0 @@
...
---   @'notAuthenticated'@ instead of redirecting to login.
-requireAuth_ :: Handler (Entity User)
-requireAuth_ = fromMaybeM notAuthenticated maybeAuth
-
-requireMemberSite :: SiteId -> Handler Site
-requireMemberSite siteId = do
-    userId <- requireAuthId
-    fromMaybe404 $ runDB $ findMemberSite siteId userId
-
--- N.B. This should be equivalent to liftM fromMaybe, but that form is
```
jferris: Yay, it's nice to see this go away. It always bugged me.

jferris:
This looks good.

PR #190, opened 2015-03-19
Extract, improve, and test authenticateUser

pbrisbin:
This code has actually been a source of bugs. With the new scaffold's
`Import.NoFoundation`, we're able to extract more code out of `Foundation`
without running into circular imports. Moving this to `Model.User` made it easy
to test.

### DIFF
--- a/Foundation.hs
+++ b/Foundation.hs
@@ -8,6 +8,7 @@ import Yesod.Default.Util   (addStaticContentExternal)
 import Yesod.Core.Types     (Logger)
 import qualified Yesod.Core.Unsafe as Unsafe
 
+import Model.User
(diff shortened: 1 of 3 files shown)


jferris:
This looks like an improvement to me. Did the behavior change here, or is this just a refactoring?

Nice to know about `Import.NoFoundation`.

pbrisbin:
Nope, no behavior change. I wanted to refactor a bit as I pulled this example into a blog post, then I noticed I could get some test coverage on it first.

pbrisbin:
This'll be useful when we introduce other authentication plugins, to be able to test the logic of handling `credsExtra` from multiple sources.

PR #192, opened 2015-03-19
Set up the expectation of a paid service

pbrisbin:


templates/docs.hamlet
```diff
@@ -142,12 +142,20 @@
   <h2>Support
 
   <p>
+
+    The Carnival hosted service is currently in Beta. We'd love for you to sign
+    up and give it a try to help us make a great product. It will eventually be
+    a paid service.
```
jferris: Do we want to add something like, "but for now, you can try it for free?"

jferris:
This seems good to me. Maybe we should get the copyediting crew to review the site in general?

pbrisbin:
Do you think anything should change here knowing that #195 is imminent?

jferris:
I think it's fine to adjust whatever we need to in #195 after this is merged.

pbrisbin:
Who's the copy-editing crew? This was suggested by @cpytel

jferris:
> Who's the copy-editing crew?

I think I've seen @gabebw, @mike-burns, @adarsh, and @croaky review copy most often. I just created a @thoughtbot/copy-editing team for this kind of thing.

gabebw:
This looks reasonable.

PR #195, opened 2015-03-20
Add Plans page with pricing information

pbrisbin:
Very rough draft. Needs copy and design.

### DIFF
--- a/Model/User.hs
+++ b/Model/User.hs
@@ -4,9 +4,11 @@ module Model.User
     , userGravatar
     , findUsers
     , findUsers'
+    , disableCreateSite
     ) where
 
 import Import
+import Plan
 
 import Network.Gravatar
 
@@ -29,3 +31,6 @@ findUsers userIds = selectList [UserId <-. userIds] []
 -- | Same as @findUsers@, but discards the @entityKey@
 findUsers' :: [UserId] -> DB [User]
 findUsers' = fmap (map entityVal) . findUsers
+
+disableCreateSite :: Int -> User -> Bool
+disableCreateSite siteCount = (>= siteCount) . planSiteCount . userPlan
--- a/Model/User.hs
+++ b/Model/User.hs
@@ -33,4 +33,4 @@ findUsers' :: [UserId] -> DB [User]
 findUsers' = fmap (map entityVal) . findUsers
 
 disableCreateSite :: Int -> User -> Bool
-disableCreateSite siteCount = (>= siteCount) . planSiteCount . userPlan
+disableCreateSite siteCount = (siteCount >=) . planSiteCount . userPlan
--- a/Model/User.hs
+++ b/Model/User.hs
@@ -4,7 +4,7 @@ module Model.User
     , userGravatar
     , findUsers
     , findUsers'
-    , disableCreateSite
+    , reachedSiteQuota
     ) where
 
 import Import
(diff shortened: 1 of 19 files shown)


pbrisbin:
Some screenshots:

carnival-plans

carnival-site-limit

jferris:
I feel like we probably want to have plans go in the database. We're almost certainly going to want them in there eventually so that we don't need to deploy to make changes to third party IDs and such, so it seems like it's better to build them out that way to begin with.

templates/plans/index.hamlet
```diff
@@ -0,0 +1,13 @@
...
+    <div .panel.panel-default>
+      <div .panel-heading>
+        <h3 .panel-title>#{show plan}
+      <div .panel-body>
+        ^{planFeatures plan}
+
+<h2>Why Pay?
+
+<p>Carnival will never contain ads
+<p>Carnival will never track your behavior
```
jferris: This is sort of a lofty claim. Does this prevent us from using Intercom or Mixpanel? We might want to be more specific about what we won't do here.

templates/plans/index.hamlet
```diff
@@ -0,0 +1,13 @@
...
+<p>Carnival will always be free for personal use
```
jferris: Are we sure we want to promise this?

pbrisbin: I mean, these things are important to *me*. I see the value in hedging, but I also see the value in making a strong claim and sticking to it.

templates/plans/index.hamlet
pbrisbin: Yeah, I suppose this has to be qualified but I don't know how to say it. The big value-add over Disqus is that we don't collect / monetize on user behavior, so I want to say that -- but clearly we're going to be tracking usage statistics for our own business metrics, so where's the line?

Model/User.hs
```diff
@@ -29,3 +31,6 @@ findUsers userIds = selectList [UserId <-. userIds] []
 -- | Same as @findUsers@, but discards the @entityKey@
 findUsers' :: [UserId] -> DB [User]
 findUsers' = fmap (map entityVal) . findUsers
+
+disableCreateSite :: Int -> User -> Bool
```
jferris: Maybe `hasReachedSiteQuota` or something so that the `Int` argument here is more obvious?

pbrisbin:
> I feel like we probably want to have plans go in the database. We're almost certainly going to want them in there eventually ...

Is it going to be any harder to move them to the DB when we need that vs doing it now?

Model/User.hs
pbrisbin: Yeah, good call

templates/plans/index.hamlet
pbrisbin: What about changing all the "will never/always" to "doesn't"/"is"

jferris:
> Is it going to be any harder to move them to the DB when we need that vs doing it now?

I feel like the more behavior we build around this, the harder it will be to change later. If you think it's significantly easier to bootstrap this way, I think that's alright; I just don't want to build too much around an assumption that we know won't hold true for very long.

templates/plans/index.hamlet
jferris: Maybe instead of a bulleted list, we should have a paragraph explaining the rationale of having a paid subscription to directly support the service rather than resorting to ads and shifty behavior to monetize after the fact?

pbrisbin:
I'm going to close this and re-open as a new PR with `Plan`s actually in the DB as well the initial Stripe integration to charge for them.

PR #197, opened 2015-03-27
Include Accept header in getUser request

pbrisbin:
Now that we rely on Yesod's automatic JSON responses to authentication routes,
we must supply a proper Accept header to get them.

### DIFF
--- a/templates/embed/Carnival.coffee
+++ b/templates/embed/Carnival.coffee
@@ -56,6 +56,7 @@ class Carnival
     request = new XMLHttpRequest
     request.withCredentials = true
     request.open('GET', '@{UserR}', false)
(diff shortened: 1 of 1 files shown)


pbrisbin:
Noticed on Staging

jferris:
Should we also set this for other JSON XMLHttpRequests?

pbrisbin:
Within `Carnival.get` probably, because we only use that to pull in JSON comments. `Carnival.post` though is used for both Logging in (where we want an HTML response) and creating comments (where we want a JSON response), so I'd need to refactor it before I could do such a thing.

I elected to only touch the one that's currently buggy for this PR.

jferris:
Sounds good. Merge at will.

PR #200, opened 2015-04-01
Add Stripe Plans

pbrisbin:
- Add Plan model and Plans route
- Add Handlers to Purchase or Cancel
- Migrate Personal and Business plan records


.ghci
```diff
@@ -19,3 +19,4 @@
 
 :set -XFlexibleInstances
 :set -XRankNTypes
+:set -XStandaloneDeriving
```
JoelQ: What does this do?

pbrisbin: Let's you do this: https://github.com/thoughtbot/carnival/pull/200/files#diff-dbe3d0b943e814dec632bd166fbce775R9

Handler/Cancel.hs
```diff
@@ -0,0 +1,22 @@
+module Handler.Cancel where
+
+import Import
+import Stripe
+
+postCancelR :: Handler Html
+postCancelR = do
+    plan <- runDB $ getBy404 $ UniquePlan freePlanId
```
JoelQ: Why `getBy404` ? Does this refer to the HTTP 404 status?

Handler/Cancel.hs
```diff
@@ -0,0 +1,22 @@
...
+    Entity userId user <- requireAuth
```
JoelQ: Does bind `requireAuth` to `user` or `Entity userId user` ?

Handler/Cancel.hs
```diff
@@ -0,0 +1,22 @@
...
+
+    result <- runStripe $ cancelSubscription $ userStripeId user
+
+    case result of
+        Just _ -> do
+            runDB $ update userId [UserPlan =. entityKey plan]
+
+            setMessage "Plan changed!"
```
JoelQ: Should this say "Plan canceled" ? Does carnival use I18n ?

Handler/Cancel.hs
```diff
@@ -0,0 +1,22 @@
...
+            redirect RootR
+
+        _ -> do
```
JoelQ: if `result` is not a `Just _`, the only other value it could have is `Nothing` right? Should that be used here instead?

Handler/Plan.hs
```diff
@@ -0,0 +1,22 @@
+module Handler.Plan where
+
+import Import
+
+getPlansR :: Handler Html
+getPlansR = do
+    plans <- runDB $ selectList [] [Asc PlanSort]
+    Entity _ user <- requireAuth
```
JoelQ: I'm confused by the `_` here. Can you call the `Entity` constructor and leave off an argument?

Handler/Plan.hs
```diff
@@ -0,0 +1,22 @@
+module Handler.Plan where
+
+import Import
+
+getPlansR :: Handler Html
+getPlansR = do
+    plans <- runDB $ selectList [] [Asc PlanSort]
```
JoelQ: What does `selectList [] [Asc PlanSort]` do?

Handler/Cancel.hs
pbrisbin: It does. The normal `get` and `getBy` functiosn return `Maybe`, while `get404` and `getBy404` don't, but short-circuit the handler with a 404 status when the record's not found.

https://www.fpcomplete.com/hoogle?q=getBy404&env=ghc-7.8-stable-14.09

Handler/Cancel.hs
pbrisbin: The left hand side is pattern matching. `requireAuth` returns an `Entity User`, which is the `UserId` and `User` together in one data type. This just pattern matches it into separate variables because they're used separately after.

test/Factories.hs
```diff
@@ -27,13 +21,28 @@ buildSite = Site
     , siteLanguage = "en-us"
     }
 
+createFreePlan :: DB (Entity Plan)
+createFreePlan = upsert Plan
```
jferris: Looks like `upsert` with no update for existing records is a `repsert`.

Handler/Cancel.hs
pbrisbin: The UI terminology is slightly off from the backend. To the user they're *changing* from Business to Personal. To us, we're *cancelling* Business.

> Does carnival use I18n?

Not yet, but that's an enhancement I'd like to make some day.

jferris:
This looks good to me. I had one minor comment, but I think you can merge this whenever Travis cheers up.

Model/User.hs
```diff
@@ -44,9 +45,11 @@ upsertUser user = entityKey <$> upsert user
...
-credsToUser :: Creds m -> Maybe User
-credsToUser Creds{..} = User
+credsToUser :: Maybe PlanId -> Creds m -> Maybe User
+credsToUser mplanId Creds{..} = User
     <$> lookup "name" credsExtra
     <*> lookup "email" credsExtra
     <*> pure credsPlugin
     <*> pure credsIdent
+    <*> mplanId
+    <*> pure Nothing
```
JoelQ: Is `pure` a more generic version of `liftIO` ?

Handler/Plan.hs
pbrisbin: This is not calling the `Entity` constructor. It's pattern matching on `Entity`. Since we don't care about the `UserId` field, we ignore it with a wildcard pattern (`_`).

Handler/Cancel.hs
pbrisbin: It's common to just use a catch-all wildcard even in cases where all but one pattern has already been matched.

Stripe.hs
```diff
@@ -0,0 +1,46 @@
...
+
+    void $ S.createSubscription stripeId planId []
+
+    return stripeId
+
+cancelSubscription :: Maybe S.CustomerId -> S.Stripe ()
+cancelSubscription Nothing = return ()
+cancelSubscription (Just stripeId) = do
+    subs <- S.list <$>
+        S.getSubscriptions stripeId (Just 1) Nothing Nothing
```
JoelQ: What does this hardcoded `1` mean?

Handler/Plan.hs
pbrisbin: https://www.fpcomplete.com/hoogle?q=selectList&env=ghc-7.8-stable-14.09

It selects all plans matching no filters (first argument), and sorting by `PlanSort` ascending (second argument).

.ghci
JoelQ: So this is like a project-wide pragma?

Handler/Cancel.hs
JoelQ: Ah, clever!

Handler/Cancel.hs
JoelQ: :+1:

Handler/Cancel.hs
JoelQ: :+1:

Handler/Plan.hs
JoelQ: Ah, makes sense

test/Factories.hs
pbrisbin: I think the docs are wrong. `repsert` requires a `PlanId` to know what to replace. `upsert` can take a `Plan` and update by it's (only) unique constraint.

Model/User.hs
pbrisbin: Nope, totally unrelated.

https://www.haskell.org/hoogle/?hoogle=pure

`pure` is from `Applicative`. It puts a value in a default, minimal context: `Applicative f => a -> f a`. In this case, the `f` is `Maybe` and it's implementation of `pure` is `Just`. Since the type of that expression needs to be `Maybe (Maybe S.CustomerId)`, `pure Nothing` works.

https://www.haskell.org/hoogle/?hoogle=liftIO

`liftIO` on the other hand takes an `IO` action and allows you to perform it in any monad that has a `MonadIO` instance: `liftIO :: MonadIO m => IO a -> m a`. It's common to do that in `Yesod` apps because you may want to perform `IO` from within `Handler` or `Widget`. Both have `MonadIO` instances, so `liftIO` works.

Stripe.hs
pbrisbin: http://hackage.haskell.org/package/stripe-haskell-0.1.3.1/docs/Web-Stripe-ApplicationFee.html#t:Limit

It's the `Limit`. Just says return, at most `1` subscriptions. We expect users to only have 1 at any given time.

.ghci
pbrisbin: This specific line is `ghci` config. It enables the extension when invoking the interpreter (common in my workflow). It was also added to the `default-extensions` field of the cabal file, enabling it any time cabal builds the project -- that's more of the "project-wide pragma", I think.

JoelQ:
LGTM once CI is green :rocket:

Handler/Purchase.hs
```diff
@@ -0,0 +1,28 @@
...
+            runDB $ update userId
+                [ UserPlan =. planId
+                , UserStripeId =. Just stripeId
+                ]
+            setMessage "Plan changed!"
+            redirect RootR
+
+        _ -> do
+            setMessage "Oops! We ran into an error, please try again."
+            redirect $ PlansR
```
sgrif: Is it worth pulling this, and the other failure case into a `genericFailure` function?

pbrisbin:
So. Here's what seemed to be happening with the builds:

- Travis was flaking due to out of memory errors
- We tried a Circle build from a different branch to see if it had the same issues (so far it doesn't)
- I tried a Circle build on this branch, and got Green
- Now that Travis can use a cached sandbox (built successfully on Circle), it doesn't fail for out of memory but does encounter a shared library error because the sandbox built on Circle is not compatible with Travis in a very subtle way (I don't know all the details here, but can dig/explain if we think that's valuable -- I do not).

I suggest the following:

- Move this project to CircleCI: cherry-pick the `circle.yml` into master (and remove `travis.yml`)
- Rebase, Green, Merge, Deploy

Handler/Purchase.hs
pbrisbin: I'm leaning towards not. To do that, I'd have to decide which module to put it in (or make a new one, then decide what to name it) -- and nothing springs to mind. I'm going to cite the Rule of Three and consider this again on a third occurrence.

jferris: I agree with leaving it alone for now, since we're still figuring out what the ideal way is to handle this kind of error.

PR #202, opened 2015-04-02
Replace Travis with CircleCI

pbrisbin:


jferris:
My build failed because of missing environment variables: https://circleci.com/gh/thoughtbot/carnival/14

Did you add those to the project configuration?

pbrisbin:
I added them in the branch where they're required, to see that it would pass. Once I merge this and rebase that, that branch will *only* add them.

PR #204, opened 2015-04-02
Disable Create when over quota, remove Beta/Free copy

pbrisbin:
The landing page is a little sparse now, but I'm not the best copy writer...

### DIFF
--- a/Handler/Sites.hs
+++ b/Handler/Sites.hs
@@ -9,10 +9,11 @@ import Yesod.Form.Bootstrap3
 
 getSitesR :: Handler Html
 getSitesR = do
-    userId <- requireAuthId
+    Entity userId user <- requireAuth
 
     (widget, enctype) <- generateFormPost $ siteForm Nothing
 
+    plan <- runDB $ belongsToJust userPlan user
     sites <- runDB $ findSites userId
 
     defaultLayout $ do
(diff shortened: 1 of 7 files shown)


Handler/Sites.hs
```diff
@@ -31,6 +32,7 @@ postSitesR = do
         setMessage "Site created!"
         redirect $ SiteR siteId
 
+    plan <- runDB $ belongsToJust userPlan user
```
jferris: This is cool.

Handler/Sites.hs
```diff
@@ -22,7 +23,7 @@ getSitesR = do
 
 postSitesR :: Handler Html
 postSitesR = do
-    userId <- requireAuthId
+    Entity userId user <- requireAuth
```
jferris: Do we want to add a quota check here to prevent race conditions or people hacking the form?

pbrisbin: I don't care about people hacking the form, myself. The effort required to craft a request (with an authenticated session) is enough of a deterrent IMO. Could be worth, at some point, tracking if enough people are curiously over quota to warrant locking down.

If I understand it correctly, there's no race condition now -- but there would be, if we added the check here. Is that right?

jferris: If you submit the form from two different sessions at once, you'll create two sites and the button will never be disabled.

jferris: > The effort required to craft a request

I would personally just use the web inspector to enable the button.

pbrisbin: Ha true.

pbrisbin: Done.

Model/Site.hs
```diff
@@ -38,3 +39,6 @@ destroySite siteId = do
     deleteWhere [CommentSite ==. siteId]
     deleteWhere [MembershipSite ==. siteId]
     delete siteId
+
+overSiteQuota :: Int -> Plan -> Bool
```
jferris: Looks like this is always called with `length sites` - thoughts on changing the type to `[a] -> Plan Bool`?

pbrisbin: Sure, why not.

jferris:
I had a suggestion about the `overSiteQuota` function, but I think this is ready to merge.

I think there's a lot we can do to improve our test factories, although I think this isn't the place. For example, if `createSite` could generate valid, unique sites, you could just use `replaceM_`.

PR #205, opened 2015-04-02
Take advantage of AuthenticationResult

pbrisbin:
- Modify authenticateUser to return an AuthenticationResult
- Internally, use Either to propagate error messages that will
 ultimately get logged in the case of failed authentications
- Specifically parse credsExtra based on the plugin used
- Separate plugin-specific behavior in implementation and test

Reasoning:

Using AuthenticationResult and returning useful ServerError values, should help
debugging. Splitting out the credsExtra parsing this way will making adding new
plugins (that return different credsExtra maps) easier. Supporting the dummy
plugin via dummyProfile makes it possible to write plugin-agnostic examples in
the test and only test plugin-specific behavior in the GitHub (and eventually
Google, Twitter, etc) contexts.

### DIFF
--- a/Model/User.hs
+++ b/Model/User.hs
@@ -13,6 +13,11 @@ import Network.Gravatar
 
 import qualified Data.Text as T
 
+data Profile = Profile
+    { profileName :: Text
+    , profileEmail :: Text
+    }
+
 instance ToJSON (Entity User) where
     toJSON (Entity uid u) = object
         [ "id" .= String (toPathPiece uid)
@@ -31,13 +36,19 @@ findUsers userIds = selectList [UserId <-. userIds] []
 findUsers' :: [UserId] -> DB [User]
 findUsers' = fmap (map entityVal) . findUsers
 
-authenticateUser :: Creds m -> DB (Maybe UserId)
+authenticateUser :: AuthId m ~ UserId => Creds m -> DB (AuthenticationResult m)
 authenticateUser creds@Creds{..} = do
-    plan <- getBy $ UniquePlan freePlanId
+    plan <- getEither "Free plan not found" $ UniquePlan freePlanId
     muser <- getBy $ UniqueUser credsPlugin credsIdent
-    muserId <- mapM upsertUser $ credsToUser (entityKey <$> plan) creds
+    euserId <- mapM upsertUser $ credsToUser (entityKey <$> plan) creds
 
-    return $ muserId <|> (entityKey <$> muser)
+    return $ case (muser, euserId) of
+        (Just user, _) -> Authenticated $ entityKey user
+        (_, Right userId) -> Authenticated userId
+        (_, Left err) -> ServerError $ credsPlugin ++ ": " ++ err
+
+  where
+    getEither msg = fmap (maybe (Left msg) Right) . getBy
 
 upsertUser :: User -> DB UserId
 upsertUser user = entityKey <$> upsert user
@@ -45,11 +56,31 @@ upsertUser user = entityKey <$> upsert user
     , UserEmail =. userEmail user
     ]
 
-credsToUser :: Maybe PlanId -> Creds m -> Maybe User
-credsToUser mplanId Creds{..} = User
-    <$> lookup "name" credsExtra
-    <*> lookup "email" credsExtra
+credsToUser :: Either Text PlanId -> Creds m -> Either Text User
+credsToUser eplanId Creds{..} = User
+    <$> (profileName <$> eprofile)
+    <*> (profileEmail <$> eprofile)
     <*> pure credsPlugin
     <*> pure credsIdent
-    <*> mplanId
+    <*> eplanId
     <*> pure Nothing
+
+  where
+    eprofile = extraToProfile credsPlugin credsExtra
+
+extraToProfile :: Text -> [(Text, Text)] -> Either Text Profile
+extraToProfile "dummy" _ = Right dummyProfile
+extraToProfile "github" extra = githubProfile extra
+extraToProfile plugin _ = Left $ "Invalid plugin: " ++ plugin
+
+dummyProfile :: Profile
+dummyProfile = Profile "Dummy Login" "dummy@example.com"
+
+githubProfile :: [(Text, Text)] -> Either Text Profile
+githubProfile extra = Profile
+    <$> lookupExtra "name" extra
+    <*> lookupExtra "email" extra
+
+lookupExtra :: Text -> [(Text, b)] -> Either Text b
+lookupExtra k extra =
+    maybe (Left $ "missing key " ++ k) Right $ lookup k extra
--- a/test/Factories.hs
+++ b/test/Factories.hs
@@ -37,8 +37,8 @@ createUser ident = do
     Entity planId _ <- createFreePlan
 
     insertEntity User
-        { userName = "John Smith (" ++ ident ++ ")"
-        , userEmail = "john-" ++ ident ++ "@gmail.com"
+        { userName = "Dummy Login"
+        , userEmail = "dummy@example.com"
         , userPlugin = "dummy"
         , userIdent = ident
         , userPlan = planId
--- a/test/Model/UserSpec.hs
+++ b/test/Model/UserSpec.hs
@@ -5,7 +5,7 @@ module Model.UserSpec
 
 import TestImport
 import Model.User
-import Yesod.Auth (Creds(..))
+import Yesod.Auth (AuthenticationResult(..), Creds(..))
 import qualified Database.Persist as DB
 
 main :: IO ()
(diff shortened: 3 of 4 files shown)


test/Factories.hs
```diff
@@ -37,8 +37,8 @@ createUser ident = do
     Entity planId _ <- createFreePlan
 
     insertEntity User
-        { userName = "John Smith (" ++ ident ++ ")"
-        , userEmail = "john-" ++ ident ++ "@gmail.com"
+        { userName = "Dummy Login"
+        , userEmail = "dummy@example.com"
```
pbrisbin: This makes me nervous. If these values don't match those in `Model.User.dummyProfile`, tests will fail. However, I don't want to export `Profile(..)` and `dummyProfile` from `Model.User` for re-use. OK trade-off?

pbrisbin: Actually, returning `Left` here will keep the same semantics as before: dummy logins are only supported for existing users, the error will be discarded in that case.

Model/User.hs
```diff
@@ -45,11 +53,32 @@ upsertUser user = entityKey <$> upsert user
...
+    <$> (profileName <$> eprofile)
+    <*> (profileEmail <$> eprofile)
     <*> pure credsPlugin
     <*> pure credsIdent
-    <*> mplanId
+    <*> eplanId
     <*> pure Nothing
+
+  where
+    eplanId = maybe (Left "Free plan not found") Right mplanId
```
jferris: I wonder if it would be better to do the `Maybe` to `Either` conversion in `authenticateUser`, since this function doesn't actually look for the free plan.

Model/User.hs
```diff
@@ -31,13 +36,16 @@ findUsers userIds = selectList [UserId <-. userIds] []
...
     plan <- getBy $ UniquePlan freePlanId
     muser <- getBy $ UniqueUser credsPlugin credsIdent
-    muserId <- mapM upsertUser $ credsToUser (entityKey <$> plan) creds
+    euserId <- mapM upsertUser $ credsToUser (entityKey <$> plan) creds
 
-    return $ muserId <|> (entityKey <$> muser)
+    return $ case (muser, euserId) of
+        (Just user, _) -> Authenticated $ entityKey user
+        (_, Right userId) -> Authenticated userId
+        (_, Left err) -> ServerError $ pack err
```
jferris: Can we use `Either Text b` in these functions and then avoid the `pack` here?

test/Factories.hs
jferris: Does it actually matter what these values are? It seems like using the dummy plugin will result in the dummy profile, regardless of the provided login details.

Model/User.hs
pbrisbin: This was factored out of a later branch that makes use of `Data.Aeson.eitherDecode :: Either String a` to parse the Google `credsExtra` values, so I'd just have to put at somewhere else.

Model/User.hs
pbrisbin: Sure, seems reasonable.

test/Model/UserSpec.hs
```diff
@@ -14,61 +14,101 @@ main = hspec spec
...
+            it "errors if email is missing" $ do
+                let creds = Creds
+                        { credsPlugin = "github"
+                        , credsIdent = "1"
+                        , credsExtra = [("name", "foo")]
+                        }
+
+                tryAuthentication creds `shouldReturn` Left "missing key email"
+
+-- Helper required to disambiguate AuthId m ~ UserId type
```
jferris: I don't understand this helper.

test/Factories.hs
pbrisbin: > It seems like using the dummy plugin will result in the dummy profile, regardless of the provided login details.

There are no "provided login details". When you login via Dummy you only specify ident, not name/email or anything like that. The issue is that any values already on an existing user are overwritten.

> Does it actually matter what these values are?

It matters that it matches what's on the `dummyProfile` side. This is because in the `Handler.Comment` spec, we assert on a response that includes user name/email. Therefore, the factory user here (used as expected) has to have values matching the `dummyProfile` (which ultimately ends up on actual)

Does that make sense?

test/Model/UserSpec.hs
pbrisbin: This may have been a mistake on my part, I'll check.

In the meantime, I can explain a little...

It has to do with type families: `AuthId m ~ UserId` means `AuthId m` and `UserId` are in a type family. It's how `Yesod.Auth` can have code that relies on `UserId` without actually needing to know about your `UserId` (i.e. `Key User`) type. It writes code using `AuthId m`, then you define the type family relationship yourself (e.g. [here](https://github.com/thoughtbot/carnival/blob/master/Foundation.hs#L130)).

When I did:

```haskell
Authenticated userId <- authenticateUser creds

Just user <- runDB $ DB.get userId
```

I got a type error because `userId` is `AuthId m` and `DB.get` requires `UserId`. My extracting the helper that has a concrete `UserId` return type, it made the compiler happy.

But, now that I think about it, it might be the ambiguous `m` that's giving trouble, not the type familiar. I.e. if I fix that to `App` in `Model.User`, it might work without this helper.

Model/User.hs
jferris: Hmm, okay. I feel like it would be good to lean towards `Text` instead of `String` in cases where both are in play, but I don't think this matters much here.

test/Factories.hs
jferris: Yeah. I think there is something funny about the data coupling here, but I think this is good to merge. Once we've played around more with testing and authentication, we may think of something better.

test/Model/UserSpec.hs
pbrisbin: Yes, so the helper is required to fix `m` as `App` because `AuthId m` is too generic to be used as a `UserId` (which is `AuthId App`). It has to be generic for... reasons:

- Fixing it to `UserId` doesn't work b/c `AuthId App ~ UserId` happens in Foundation
- Fixing it to `AuthId App` doesn't work b/c `App` is defined in Foundation
- I can't import Foundation in Model.User b/c that's a cycle

So... a helper or explicit annotation is needed, but I can at least clarify the comment

Model/User.hs
```diff
@@ -31,13 +36,16 @@ findUsers userIds = selectList [UserId <-. userIds] []
 findUsers' :: [UserId] -> DB [User]
 findUsers' = fmap (map entityVal) . findUsers
 
-authenticateUser :: Creds m -> DB (Maybe UserId)
+authenticateUser :: AuthId m ~ UserId => Creds m -> DB (AuthenticationResult m)
```
JoelQ: What is the `~` operator?

test/Factories.hs
pbrisbin: > I think there is something funny about the data coupling here

Definitely, I think any time you talk about a "dummy" login in your app, funny things may happen. The real fix would be an less hacky way to authenticate in tests.

Model/User.hs
JoelQ: variables prefixed with `m` are `Maybe`s and variables prefixed with `e` are `Either`s ?

Model/User.hs
```diff
@@ -45,11 +53,32 @@ upsertUser user = entityKey <$> upsert user
...
-    <*> mplanId
+    <*> eplanId
     <*> pure Nothing
+
+  where
+    eplanId = maybe (Left "Free plan not found") Right mplanId
+    eprofile = extraToProfile credsPlugin credsExtra
+
+extraToProfile :: Text -> [(Text, Text)] -> Either String Profile
+extraToProfile "dummy" _ = Right $ dummyProfile
```
JoelQ: Do you need the `$` here? Can this just be `Right dummyProfile` ?

Model/User.hs
```diff
@@ -45,11 +53,32 @@ upsertUser user = entityKey <$> upsert user
...
+extraToProfile "github" extra = githubProfile extra
+extraToProfile plugin _ = Left $ "Invalid plugin: " ++ unpack plugin
+
+dummyProfile :: Profile
+dummyProfile = Profile "Dummy Login" "dummy@example.com"
+
+githubProfile :: [(Text, Text)] -> Either String Profile
+githubProfile extra = Profile
```
JoelQ: Does this need to wrap `Profile` in `Right` ?

Model/User.hs
pbrisbin: That's the plan, yup.

Model/User.hs
pbrisbin: It has to do with type families, see this comment: https://github.com/thoughtbot/carnival/pull/205#discussion_r27682987

Model/User.hs
pbrisbin: Yup, good catch. I'll `hlint` the whole file and address anything found.

Model/User.hs
pbrisbin: Nope, this is an Applicative expression. `Profile` is combined with the two lookups via `(<$>)` and `(<*>)`. Because of the semantics of those operators for `Either e`, if both are found you get a `Right profile` otherwise a `Left error`.

pbrisbin: Also, I do compile before committing ;) And if this indeed required a `Right` it would be a type error the way it is.

JoelQ: :+1: This was more of a "I'm not sure what is going on here" question than a "do you even compile?!"

JoelQ:
LGTM :+1:

pbrisbin:
Addressed all comments, rebased, will merge on Green. Thanks!

PR #207, opened 2015-04-02
Improvements to deploy procedure

pbrisbin:
- Use Heroku Pipelines, not hand-rolled copy-slug script
- Pass BUILD=1 to automate Halcyon build steps


bin/deploy
```diff
@@ -4,13 +4,24 @@
...
 set -e
 
 case "${1:-carnival-staging}" in
-  carnival-staging) git push staging master ;;
+  carnival-staging)
+    git push staging master
+
+    if [ "$BUILD" = 1 ]; then
+      heroku run --size=PX build
+      git commit --allow-empty -m 'Trigger deployment'
```
jferris: Can we do this on a branch so that we don't pollute master with empty commits?

bin/deploy
```diff
@@ -4,13 +4,24 @@
...
 #
+#   Set BUILD=1 to perform a Halcyon build
+#
 ###
 set -e
 
 case "${1:-carnival-staging}" in
-  carnival-staging) git push staging master ;;
+  carnival-staging)
+    git push staging master
```
jferris: Think it's worth doing a `-f` here?

bin/deploy
pbrisbin: Can you elaborate how this would work?

bin/deploy
pbrisbin: When would that be needed?

pbrisbin:
Ping. I've almost accidentally used copy-slug twice today...

Maybe we can take up the empty-commit question separately? It's how I've been doing deploys all along so this PR doesn't really make it worse -- just automates it.

bin/deploy
jferris:
``` sh
git checkout -b deploy
git commit --allow-empty -m "Trigger deployment"
git push -f staging deploy:master
git checkout master
git branch -D deploy
```

bin/deploy
jferris: It's largely related to the "empty commit on a branch" thing, since the empty commit will be deployed to staging but not master.

I have a strong, negative reaction to having artifacts of the deploy environment in Git history. Things like empty commits and "skipci" tags all really bug me.

bin/deploy
pbrisbin: So that would mean that all deploys to staging (building or not) will have to force -- right? Is that OK? I get a twitch when automating anything using the word "force"

jferris:
> Maybe we can take up the empty-commit question separately?

I'm okay with this, but here's another suggestion: can we just merge the switch to pipelines without the Halcyon automation changes?

Either way, I think this is okay to merge, but my personal preference is to merge with just the pipeline switch.

pbrisbin:
Done. #211

PR #208, opened 2015-04-03
Improvements to authentication spec

pbrisbin:
- Derive Show/Eq for AuthenticationResult
- Extract shouldCreateWith/shouldUpdateWith assertions

Another change pulled out of a branch where I'm adding Google authentication.

The Show/Eq instances allow assertions against AuthenticationResult values
which replace generic pattern match failures with something useful. The
extracted assertions are reused when testing Google logins. Once Profile was
exported for the assertions, it made sense to also export dummyProfile to
remove that data duplication.

### DIFF
--- a/Model/User.hs
+++ b/Model/User.hs
@@ -5,6 +5,10 @@ module Model.User
     , findUsers
     , findUsers'
     , authenticateUser
+
+    -- Exported for tests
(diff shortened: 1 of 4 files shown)


jferris:
This looks good to me.

PR #210, opened 2015-04-03
"Powered by Carnival" branding

pbrisbin:
Desktop:

https://files.pbrisbin.com/screenshots/desktop_150403154819.png

Mobile:

https://files.pbrisbin.com/screenshots/desktop_150403154833.png

### DIFF
--- a/Model/Site.hs
+++ b/Model/Site.hs
@@ -4,9 +4,13 @@ module Model.Site
     , createSite
     , destroySite
     , overSiteQuota
+    , siteBranded
     ) where
 
 import Import
+import Database.Esqueleto hiding ((==.), delete, on)
+
+import qualified Database.Esqueleto as E
 
 findSites :: UserId -> DB [Entity Site]
 findSites userId = do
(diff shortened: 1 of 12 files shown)


jferris:
The screen shots aren't working for me.

Model/Site.hs
```diff
@@ -4,9 +4,13 @@ module Model.Site
     , createSite
     , destroySite
     , overSiteQuota
+    , siteBranded
     ) where
 
-import Import
+import Import hiding (on)
+import Database.Esqueleto hiding ((==.), delete)
```
jferris: Out of curiosity, what is `==.` currently defined from?

Model/Site.hs
```diff
@@ -42,3 +46,20 @@ destroySite siteId = do
 
 overSiteQuota :: [a] -> Plan -> Bool
 overSiteQuota xs Plan{..} = length xs >= planSiteQuota
+
+siteBranded :: SiteId -> DB Bool
+siteBranded siteId = do
```
jferris: Given a site's ID, it seems like we should be able to answer the question in a single SQL query.

pbrisbin: Oh right. I had originally split it because I wanted to handle no members specially, but I'm not doing that any longer. Back down the esqueleto rabbit hole...

pbrisbin:
Screenshots might've been a GitHub problem -- just put links for now.

Model/Site.hs
pbrisbin: `Database.Persistent`, Esqueleto overrides many of them to keep usage familiar.

Model/Site.hs
```diff
@@ -4,9 +4,13 @@ module Model.Site
     , createSite
     , destroySite
     , overSiteQuota
+    , siteBranded
     ) where
 
-import Import
+import Import hiding (on)
```
jferris: I think it's easier to keep the imports in your head if we're consistent with hiding. In this case, I think using `E.on` is more obvious. That way, everything unqualified is from `Import`.

jferris:
The screen shots look good to me.

pbrisbin:
Updated to use single select and `E.on`

jferris:
Looks good to merge.

PR #211, opened 2015-04-03
Automate Halcyon steps if BUILD=1

pbrisbin:
- Using the environment variable (vs a flag to bin/deploy) allows for handling
  builds on migrate-deploys too, where you don't invoke bin/deploy directly.

### DIFF
--- a/bin/deploy
+++ b/bin/deploy
@@ -4,11 +4,27 @@
 #
 #   APP defaults to carnival-staging
 #
+#   Set BUILD=1 to perform a Halcyon build
+#
 ###
 set -e
 
 case "${1:-carnival-staging}" in
-  carnival-staging) git push staging master ;;
+  carnival-staging)
+    git push --force staging master
+
+    if [ -n "$BUILD" ]; then
+      heroku run --size=PX build
+
+      git checkout -b deploy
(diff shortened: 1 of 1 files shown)


pbrisbin:
Suggestion:

```
git checkout -b deploy
git commit --allow-empty -m "Trigger deployment"
git push -f staging deploy:master
git checkout master
git branch -D deploy
```

bin/deploy
```diff
@@ -4,11 +4,27 @@
...
+#
 ###
 set -e
 
 case "${1:-carnival-staging}" in
-  carnival-staging) git push staging master ;;
+  carnival-staging)
+    git push --force staging master
+
+    if [ "$BUILD" = 1 ]; then
```
jferris: Any reason to look for `1` and not just a non-empty value? I prefer to use `BUILD=OHYEAH`.

jferris:
I had a minor comment, but this looks good to merge whenever you're ready.

bin/deploy
pbrisbin: No reason. I just usually err on the explicit and exact side. If you want OHYEAH though, you can have it.

PR #212, opened 2015-04-03
Add GoogleEmail2 authentication

pbrisbin:
TODO:

- Create OAuth2 apps on Google for Staging/Production
- Design the Login page
- Homepage call to action should go to Login (not go direct to GH)
- Embed should open Login (not go direct to GH)


pbrisbin:
Ping @smharley for the design work on Login/Homepage

Model/User.hs
```diff
@@ -85,6 +89,28 @@ githubProfile extra = Profile
...
 
+googleProfile :: [(Text, Text)] -> Either Text Profile
+googleProfile extra = Profile
+  <$> (handleName =<< decodeText =<< lookupExtra "name" extra)
+  <*> (handleEmails =<< decodeText =<< lookupExtra "emails" extra)
+
+  where
+    handleName :: Name -> Either Text Text
+    handleName Name{..} =
+        case (nameFormatted, nameGiven, nameFamily) of
```
jferris: This is...interesting. Thanks, Google.

test/Model/UserSpec.hs
```diff
@@ -82,6 +82,56 @@ spec = withApp $ do
...
+            it "errors if name has no values" $ do
+                let creds' = creds
+                        { credsExtra =
+                            [ ("name", "{}")
+                            , ("email", "[{\"value\":\"\",\"type\":\"account\"")
+                            ]
+                        }
+
+                runDB (authenticateUser' creds')
+                    `shouldReturn` ServerError "googleemail2: user has no name"
```
jferris: Do we display these errors to the user?

pbrisbin: This is handled by the change I made in `yesod-auth`. The behavior depends on what we return:

- `Authenticated user` - all OK
- `UserError msg` - Show `msg` to the user (it's always "Invalid Login" for now)
- `ServerError err` - Log `err` and show the user "Authentication failure"

jferris: Awesome.

jferris:
This looks good from a development perspective.

pbrisbin:
Thoughts on merging this as-is?

- Keep the Homepage and Embed going direct to GitHub
- The only difference would be if someone ends up on /auth/login directly, there will be a Google option there
- Follow up separately with design work and change it so users go to login and not directly to GitHub

jferris:
That sounds like a good plan to me.

PR #213, opened 2015-04-03
Remove temporary database defaults

### LINKED ISSUE #199: Remove all temporary defaults in config/model

pbrisbin:
These were used to migrate existing records when deploying new fields. All
records in all environments are populated at this point.

Resolves #199


jferris:
Looks good!

PR #216, opened 2015-04-07
Force SSL in non-development environments

pbrisbin:
http://hackage.haskell.org/package/yesod-core-1.4.9.1/docs/Yesod-Core.html#v:sslOnlyMiddleware


jferris:
Looks good to me.

PR #217, opened 2015-04-07
Add sslOnlySessions

pbrisbin:
Forgot we need this as well...

I'm going to hold off on implementing a forced redirect for now and instead try
to add that to Yesod proper and update Carnival after.

### DIFF
--- a/Foundation.hs
+++ b/Foundation.hs
@@ -50,7 +50,7 @@ instance Yesod App where
 
     -- Store session data on the client in encrypted cookies,
     -- default session idle timeout is 120 minutes
-    makeSessionBackend _ =
+    makeSessionBackend _ = whenSSL sslOnlySessions $
         Just <$> envClientSessionBackend sessionTimeout "SESSION_KEY"
 
     defaultLayout widget = do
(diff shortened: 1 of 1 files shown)


Foundation.hs
```diff
@@ -114,11 +114,10 @@ instance Yesod App where
...
 
     yesodMiddleware =
-        forceSSL (appForceSSL compileTimeAppSettings) . defaultYesodMiddleware
+        whenSSL (sslOnlyMiddleware sessionTimeout) . defaultYesodMiddleware
 
-forceSSL :: Bool -> Handler a -> Handler a
-forceSSL True = sslOnlyMiddleware sessionTimeout
-forceSSL False = id
+whenSSL :: (a -> a) -> (a -> a)
+whenSSL f = if (appForceSSL compileTimeAppSettings) then f else id
```
jferris: I'm surprised that `compileTimeAppSettings` is available outside `Handler`, but okay!

jferris:
Looks good.

Foundation.hs
pbrisbin: Yup, it's *compileTime*. I.e. it splices the result of the yaml parse into a constant function that you can call from anywhere.

PR #226, opened 2015-04-10
Enforce unique emails on User

pbrisbin:
Caveats:

- This is only at app-level, no DB constraint (yet)
- This is only on *new* users, not profile updates to existing users (yet)

Rationale:

Adding the DB constraint fails almost all of our tests. Our user factory
actually does create users with unique emails, but authenticating with the
dummy plugin always attempts to update it to the same address. Modifying that
behavior will be both messy and unrelated to this validation, so I want to do
it separately.

I thought it made sense to do this first, since it should prevent creating
users with non-unique emails in production while I'm working on adding the more
robust, DB-level constraint.

Once the constraint is in and the test behavior corrected, I'll change the
update path to also handle uniqueness violations gracefully.

### DIFF
--- a/Foundation.hs
+++ b/Foundation.hs
@@ -11,6 +11,7 @@ import qualified Yesod.Core.Unsafe as Unsafe
 import Model.User
 import Yesod.Auth.Dummy
 import Yesod.Auth.GoogleEmail2 hiding (Token)
+import Yesod.Auth.Message (AuthMessage(..))
 import Yesod.Auth.OAuth2.Github
 
 -- | The foundation datatype for your application. This can be a good place to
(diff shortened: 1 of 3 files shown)


pbrisbin:
Adjusting the tests to allow for unique emails was easier than I thought. #226 should be merged first, then I'll rebase this (and add code for the update path).

jferris:
> #226 should be merged first, then I'll rebase this (and add code for the update path).

This is #226?

pbrisbin:
Sorry, #227, where we're having the other discussion.

pbrisbin:
Just realized: once the merge-on-existing-email logic is in place, this path would never be hit. Closing in favor of that, which I'll open soon.

PR #228, opened 2015-04-10
Handle users with the same email but another plugin

pbrisbin:
- Update the existing user record to the new plugin/ident
- Let authentication proceed from there


jferris:
Looks good.

PR #229, opened 2015-04-10
Add migration to remove duplicate emails

pbrisbin:
- Move all relations to the github user record
- Delete the learn user record

Tested locally for correct syntax. Am now going to work on creating some
example users to test that it does what's intended.

### DIFF
--- a/migrations/migrate-duplicate-emails.sql
+++ b/migrations/migrate-duplicate-emails.sql
@@ -1,0 +1,51 @@
+--------------------------------------------------------------------------------
+-- For each relation, find all records tied to a learn user that also have a
+-- github account with the same email, re-associate the records with the github
+-- account
+--------------------------------------------------------------------------------
+UPDATE membership
+  SET "user" = guser.id
+  FROM "user" luser
+  JOIN "user" guser
+    ON guser.email = luser.email
+  WHERE luser.id = membership.user
+    AND guser.plugin = 'github'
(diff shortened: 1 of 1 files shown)


migrations/migrate-duplicate-emails.sql
```diff
@@ -0,0 +1,52 @@
...
+-- github account with the same email, re-associate the records with the github
+-- account
+--------------------------------------------------------------------------------
+UPDATE membership
+  SET "user" = guser.id
+  FROM "user" luser
+  JOIN "user" guser
+    ON guser.email = luser.email
+  WHERE luser.id = membership.user
+    AND guser.plugin = 'github'
```
jferris: Will this set their `plugin` to "github" but leave a Learn UID in `ident`? Is there a potential collision issue there?

pbrisbin: This the condition clause, not the update. User records aren't changed as part of this process (at least, that's the intention). The relations (memberships, comments, and subscriptions) are re-assigned to the github user. That should leave all the (duplicate) learn records with no data associated with them, at which point they're deleted.

jferris: Ah, right - I read this query backwards. I was looking at the tables in the "FROM" clause thinking those tables were being updated.

Reading through this again, everything makes sense.

PR #231, opened 2015-04-10
Update halcyon-build script to clear cache

malandrina:
* This commit updates `bin/halcyon-build` to clear the halcyon cache
  before proceeding with the build. This change is necessary to ensure
  that archives and constraints cached over the course of previous
  failed builds are not applied when the script is run.
* References:
  - halcyon cache: https://halcyon.sh/reference/#halcyon_cache
  - an issue with some discussion around clearing the halcyon cache:
  https://github.com/mietek/halcyon/issues/22

### DIFF
--- a/bin/halcyon-build
+++ b/bin/halcyon-build
@@ -27,11 +27,6 @@ EOF
   exit 1
 fi
 
-if [ -d /var/tmp/halcyon-cache ]; then
-  echo 'Clearing the halcyon cache...'
-  rm -rf /var/tmp/halcyon-cache
-fi
-
 if [ ! -w /app ]; then
   sudo mkdir -p /app
   sudo chown $USER /app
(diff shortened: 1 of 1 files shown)


pbrisbin:
First of all, thanks for working through whatever issue you were having and trying to make it easier for others. I really appreciate it.

> This change is necessary to ensure that archives and constraints cached over the course of previous failed builds are not applied

In general, Halcyon is very good about knowing when cached artifacts aren't applicable and any cases where `--purge-cache` is needed should be considered a bug in Halcyon or due to an experienced user doing something unusual. That said, it does happen and I get that adding this is a safer default that prevents confusing errors for unfamiliar users. Unfortunately, adding this option would slow down every build in an attempt to solve a problem that occurs in very few. I don't think it's a good idea.

I'd be interested to know what error you saw that was fixed by purging the cache. Would it make more sense to add a Trouble Shooting section to DEVELOPERS that shows the error and instructs users to invoke `bin/halcyon-build --purge-cache` if/when they see it?

jferris:
@malandrina are you still having trouble setting up Carnival?

malandrina:
@pbrisbin @jferris apologies for the delay in responding. I succeeded in setting Carnival up locally a couple weeks ago.

@pbrisbin I agree it would be better not to purge the halcyon cache for local builds for the reason you mentioned. I like the idea of describing the issues I encountered in a "Troubleshooting" section and suggesting purging the cache there.

Closing this until I can review my notes on the setup process and write a summary of the issues I encountered.

PR #232, opened 2015-04-10
Add wireframe to example image

smharley:
image


smharley:
My plan is to just get something up so it isn't blank, and later do some animation to show an example

pbrisbin:
Nice! LGTM

pbrisbin:
I went ahead and merged so it will be included in the deploy I'm about to do for the login styling.

PR #233, opened 2015-04-13
Silence unused import warnings

pbrisbin:
Now that we no longer link directly to GitHub, oauth2Url is unused.


jferris:
Looks good.

PR #234, opened 2015-04-13
Add options to remove presumptive DOM changes

### LINKED ISSUE #184: Allow disabling of article left-shift
Users who require more control over what happens when comments expand should disable this and do something themselves based on the `.commenting` class we add to the article. It might also be valuable to add a `onCommenting` hook. Though, they could subscribe the the events we trigger, if we document them.

This would allow end users to solve #59 and #64 for themselves.

pbrisbin:
- Document new force_relative and shift_article options
- Make forcing position:relative conditional on option
- Make automatic shifting of article conditional on option

Resolve #184

### DIFF
--- a/templates/docs.hamlet
+++ b/templates/docs.hamlet
@@ -76,6 +76,19 @@
             <code>:scope
           will refer to your Article element.
       <tr>
+        <td><code>shift_article</code>
+        <td><code>true</code>
+        <td>
+          Automatically shift your Article element to make room for the comments
+          UI.
+      <tr>
+        <td><code>force_relative</code>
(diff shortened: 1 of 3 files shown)


jferris:
I feel like introducing this now may be premature. We don't have anybody else using the site yet, so we're not sure what people will actually want or need. I'd prefer to defer adding this functionality until we have an example of where it's needed.

pbrisbin:
I guess I didn't really look at it as functionality. Just putting an option around functionality we already have. I'll close the PR but save the branch in case we decide we want this at some point.

PR #235, opened 2015-04-13
Move close-window logic back to homepage

pbrisbin:
- Reverts 98f9c54
- Adds a "dest" param to loginHandler to ensure we redirect users to the
 root layout and fire this close-window logic (only) from embedded
 logins


Foundation.hs
```diff
@@ -171,9 +171,13 @@ instance YesodAuth App where
 
     authHttpManager = getHttpManager
 
-    loginHandler = lift $ defaultLayout $ do
-        setTitle "Carnival - Login"
-        $(widgetFile "login")
+    loginHandler = lift $ do
+        murl <- runInputGet $ iopt textField "dest"
+        maybe (return ()) setUltDest murl
```
jferris: Is `maybe` with `return ()` just `mapM_`?

jferris:
Looks good.

PR #239, opened 2015-04-14
Allow unauthenticated users to see pricing

pbrisbin:
- Replace Subscribe buttons with Sign in to subscribe links
- Redirect to login, then back to PlansR

### DIFF
--- a/templates/plans/purchase.hamlet
+++ b/templates/plans/purchase.hamlet
@@ -1,19 +1,23 @@
-$with price <- planPrice plan
-  $if userPlan user == planId
-    <form>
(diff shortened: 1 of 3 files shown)


templates/plans/purchase.hamlet
```diff
@@ -1,19 +1,23 @@
 $with price <- planPrice plan
```
jferris: Looks like `price` isn't used in the `$nothing` branch here, so you could move `$with price` inside the `$maybe`.

jferris:
I had one comment. Good to merge whenever you're ready.

pbrisbin:
Nice catch. Will merge on Green.

PR #240, opened 2015-04-14
Load Carnival on the home page

pbrisbin:


Handler/Docs.hs
```diff
@@ -1,19 +1,13 @@
 module Handler.Docs where
 
 import Import
+import Model.Site
 
 getDocsR :: Handler Html
 getDocsR = do
     root <- getAppRoot
-    Entity siteId _ <- runDB $ upsert (demoSite root) []
+    siteId <- runDB $ upsertDemoSite root
```
jferris: We could make this into a `Widget` so the function/template pair can find the `siteId` on its own, right?

pbrisbin: I'm not sure I follow, what is the template in the pair?

jferris: The JavaScript to boot up the comments.

pbrisbin: But the JavaScript is different between the two cases.

PR #245, opened 2015-04-14
Allow comments of any (non-zero) length

pbrisbin:
- Add an error message to the UI on empty comments

I'm not sure the messaging is needed once we allow comments of any non-zero
length. It's pretty obvious when submitting an *empty* comment does nothing, I
was originally confused while trying to add *non-empty* comments and it wasn't
doing anything because it was less than 4 characters (think: "+1").

That said, I added the messaging before realizing this and it's here now, so
wanted to get others' opinions before discarding it. Also, if there is a good
reason to keep the 4-character validation, I'll want the messaging again.

### DIFF
--- a/templates/embed.lucius
+++ b/templates/embed.lucius
@@ -192,6 +192,12 @@
   }
 }
 
+.carnival-validation-error {
+  color: #FF0000;
+  font-size: 12px;
+  padding-top: 10px;
+}
+
 @media only screen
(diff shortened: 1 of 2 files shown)


jferris:
I think the client-side validation is probably worth having. I can't comment on the 4 character requirement.

pbrisbin:
@cpytel what was the reasoning for the 4 character minimum?

PR #246, opened 2015-04-15
Add analytics.identify(), Secure mode support

pbrisbin:
- Identify logged-in users with Segment
- Set up Secure mode for Intercom
- Extract analytics setup to separate template
- Fix bug where `loadEnv` was not called early enough to affect
 `appSettings`

### DIFF
--- a/Foundation.hs
+++ b/Foundation.hs
@@ -134,6 +134,9 @@ embedLayout widget = do
     pc <- widgetToPageContent widget
     withUrlRenderer $(hamletFile "templates/embed-layout-wrapper.hamlet")
 
+analytics :: App -> Maybe (Entity User) -> a -> Html
+analytics app muser = $(hamletFile "templates/analytics.hamlet")
+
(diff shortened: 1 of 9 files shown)


Foundation.hs
```diff
@@ -134,6 +134,9 @@ embedLayout widget = do
     pc <- widgetToPageContent widget
     withUrlRenderer $(hamletFile "templates/embed-layout-wrapper.hamlet")
 
+analytics :: App -> Maybe (Entity User) -> url -> Html
```
jferris: `url` is lowercase in the type signature?

pbrisbin: It's a type variable. I figured I could give it *some* information, vs an opaque `a` or `t`. The argument goes unused because we don't do any URL interpolation in this template.

pbrisbin: Actually, this is the render-url function. So the name is bad anyway. Maybe just `a` would be better.

jferris:
Looks good.

PR #249, opened 2015-04-16
Fix synchronous CORS requests in Firefox 37

### LINKED ISSUE #248: Does not work in Firefox
Steps:

- Firefox 37.0.1-2
- https://carnivalapp.io
- Hover, click "Leave comment"
- See in console:

```
InvalidAccessError: A parameter or an operation is not supported by the underlying object
```

Based this: http://stackoverflow.com/questions/16677893/evil-firefox-error-a-parameter-or-an-operation-is-not-supported-by-the-under, it's because of the `async: false` on the call for user information (see #170). One solution may be here: http://stackoverflow.com/questions/16668386/cors-synchronous-requests-not-working-in-firefox

pbrisbin:
http://stackoverflow.com/questions/16668386/cors-synchronous-requests-not-working-in-firefox.

Put withCredentials = true within a beforeSend function. Resolves #248.

### DIFF
--- a/templates/embed/Carnival.coffee
+++ b/templates/embed/Carnival.coffee
@@ -20,20 +20,14 @@ class Carnival
       elem.className = newClass.replace(/^\s+|\s+$/g, '')
 
   @get: (url, callback) ->
-    request = new XMLHttpRequest
-    request.withCredentials = true
-    request.open('GET', url, true)
-    request.setRequestHeader('Content-Type', 'application/json')
+    request = @openRequest('GET', url)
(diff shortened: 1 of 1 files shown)


jferris:
The three changes and the block around them look pretty similar. Worth extracting a function?

pbrisbin:
Sure, now I'll just have to a bit more testing before being confident enough to merge. Sure wish we had something automated...

pbrisbin:
How bout now?

jferris:
Lovely!

PR #250, opened 2015-04-16
Integration testing with Test.Hspec.WebDriver

pbrisbin:
Integration testing via `Test.Hspec.WebDriver`

This is no fun to run locally. But it's nice to have on CircleCI.

### DIFF
--- a/carnival.cabal
+++ b/carnival.cabal
@@ -202,6 +202,7 @@ test-suite test
                  , case-insensitive
                  , containers
                  , hspec-expectations-lifted
+                 , hspec-webdriver
                  , http-types
                  , markdown
                  , mime-mail
@@ -212,4 +213,5 @@ test-suite test
                  , text
                  , time
                  , wai-extra
+                 , webdriver
                  , yesod-auth
--- a/test/IntegrationImport.hs
+++ b/test/IntegrationImport.hs
@@ -1,0 +1,40 @@
+module IntegrationImport
+    ( inModal
+    , clickOn
+    , fillIn
+    , waitForElem
+    , waitForText
+    , module X
+    ) where
+
+import ClassyPrelude            as X hiding (Element)
+import Test.Hspec.WebDriver     as X hiding (shouldBe, shouldReturn)
+
+import Test.WebDriver.Class
+import Test.WebDriver.Commands.Wait
+
+-- N.B. relies on the assumption that a modal has been opened and is the only
+-- window open besides the current one.
+inModal :: WebDriver m => m a -> m ()
+inModal f = do
+    cw <- getCurrentWindow
+    focusWindow =<< return . unsafeHead . filter (/= cw) =<< windows
+    void $ f
+    focusWindow cw
+
+clickOn :: WebDriver m => Selector -> m ()
+clickOn = click <=< waitForElem
+
+fillIn :: WebDriver m => Selector -> Text -> m Element
+fillIn x value = do
+    input <- findElem x
+    sendKeys value input
+    return input
+
+waitForText :: WebDriver m => Selector -> m Text
+waitForText = getText <=< waitForElem
+
+waitForElem :: WebDriver m => Selector -> m Element
+waitForElem = waitUntil timeout . findElem
+  where
+    timeout = 2000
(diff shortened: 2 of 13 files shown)


pbrisbin:
This is ready to be looked at. /cc @jferris

carnival.cabal
```diff
@@ -146,6 +146,64 @@ executable         carnival
 
     ghc-options:       -threaded -O2 -rtsopts -with-rtsopts=-N
 
+executable carnival-integration
```
jferris: Bummer that we need to repeat a lot of these options. Is there a way to specify `extension` or `ghc-options` for every target? Do we need all of these `build-depends` or are some of them only used from the library?

pbrisbin: I don't know of a way to centralize extensions, etc.

We can audit the build depends, but I suspect we need them all b/c we
import TestImport which references a lot. Might we worth it to duplicate
what we need in IntegrationImport or break up the modules differently. Then
again, the depends list isn't a huge deal...

test/IntegrationImport.hs
```diff
@@ -0,0 +1,53 @@
...
+fillIn x value = do
+    input <- findElem x
+    sendKeys value input
+    return input
+
+waitForText :: WebDriver m => Selector -> m Text
+waitForText = getText <=< waitForElem
+
+waitForElem :: WebDriver m => Selector -> m Element
+waitForElem = waitUntil 50000 . findElem
```
jferris: What is the `50000` value here? Milliseconds? Is this excessive?

jferris:
This generally looks good to me, but it's too bad how much boilerplate is required for a single integration test. I suppose some of this could be submitted or extracted to libraries.

test/IntegrationImport.hs
pbrisbin: I'm not sure, I think it's whatever threadDelay takes.

I think excessive is good, it'll only be slow in the failure case. I'd
rather that then run the risk of not waiting long enough and getting a
false failure.

pbrisbin:
@jferris I'm considering closing this as stale. The high ratio of boilerplate to value bothers me. I think I'd rather see the effort go into setting up js unit tests than trying to do something end-to-end that's janky and (probably) very brittle. WDYT?

jferris:
I think it would be valuable to have some kind of smoke test to ensure we didn't totally break the embedded API. There might be a more lightweight way to do that, though.

pbrisbin:
OK, I made a few changes and now feel much better about merging this:

- The changes needed to `TestImport` are minor (only had to extract `runApp`)
- `IntegrationImport` is much more sane (no more DB stuff, mostly just composed css manipulations for convenience)
- Clarified that the timeout in `waitForElem` is seconds (was a surprise to me) and reduced it

Ready for re-review.

jferris:
That's definitely better. I think my remaining concern is with complicating the cabal setup. I worry about having to add and manage dependencies in two places. I think there's also an issue where having multiple executables breaks `yesod devel` (see #279).

pbrisbin:
I think I can get around the duplication and second executable by using a
flag to change main-is for the test target. I'll try that out tonight or
tomorrow.

pbrisbin:
Well, I can't use a flag to conditionalize `main-is` because apparently flags can only be used in library or executable stanzas (not test-suite). But I was able to use CPP to make `Spec.hs` dynamic.

I also got CI green by increasing the wait timeout to `2000`. I'm thinking the docs are wrong and it's not actually a value in seconds, maybe milliseconds.

jferris:
Nice. This is looking really solid. The changes to the non-Selenium stuff are really minimal at this point.

If this works out for us, it looks like there's some generic stuff that could be extract to a library.

pbrisbin:
> This is looking really solid

Solid enough to merge? The only downside I see is it does make CI builds quite a bit slower (~15m -> 30/40m) because we need to recompile a lot as the flags change (which I think is unavoidable).

jferris:
> 15m -> 30/40m

Yikes! That's too bad. I think it's worth doing anyway. Maybe using something like Docker can help us out of this mess.

pbrisbin:
Rebased and updated commit message -- will merge on Green.

PR #253, opened 2015-04-17
Re-introduce openRequest duplication

pbrisbin:
As part of af142f2, the request building logic was centralized. Something
about this refactor breaks CORS requests entirely.

This commit reverts that, and introduces the Firefox fix without removing the
duplication, which we will address separately.

### DIFF
--- a/templates/embed/Carnival.coffee
+++ b/templates/embed/Carnival.coffee
@@ -20,14 +20,22 @@ class Carnival
       elem.className = newClass.replace(/^\s+|\s+$/g, '')
 
   @get: (url, callback) ->
-    request = @openRequest('GET', url)
+    request = new XMLHttpRequest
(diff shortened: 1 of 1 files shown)


gabebw:
Is this a pure `git revert af142f2`?

pbrisbin:
> This commit reverts that, and introduces the Firefox fix without removing the
duplication

gabebw:
I'll take that as a "no".

Have you confirmed this works on, say, the staging blog? Or is the idea to merge this and apply another fix if it doesn't work?

pbrisbin:
I've confirmed this works locally. Next step is staging blog, I usually merge before testing there rather than deploying a branch.

gabebw:
Reasonable. Merge away.

PR #254, opened 2015-04-17
Revert fix for Firefox

pbrisbin:
- Reverts commit 83202193f3d898c874cc1f6c14bef254755730bc.
- Reverts commit af142f2e0d4227f4cdf32ccfd36d39983cdf179e.

The fix for Firefox breaks Chrome, and didn't really fix Firefox anyway.

The reason Firefox was broken is because we make a synchronous request for
user details. We should not be doing this anyway. The decision was made to
revert the Firefox "fix" and focus on making the user request async instead.


jferris:
Cool.

PR #255, opened 2015-04-17
Adjust plans

pbrisbin:
- Rename branded to commercial
- Lower site quota on Business
- Increase, but don't show comment quotas

### DIFF
--- a/migrations/create-plan.psql
+++ b/migrations/create-plan.psql
@@ -5,6 +5,6 @@ CREATE TABLE IF NOT EXISTS plan (
   price                 INT8 NOT NULL,
   site_quota            INT8 NOT NULL,
   comment_quota         INT8 NOT NULL,
-  branded               BOOLEAN NOT NULL,
+  commercial            BOOLEAN NOT NULL,
   sort                  INT8 NOT NULL
 );
--- a/templates/plans/features.hamlet
+++ b/templates/plans/features.hamlet
@@ -1,8 +1,9 @@
 <ul>
-  <li>#{planSiteQuota plan} Site(s)
-  <li>#{planCommentQuota plan} Comments per month
-
-  $if planBranded plan
+  $if planCommercial plan
+    <li>#{planSiteQuota plan} commercial Site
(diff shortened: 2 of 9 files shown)


pbrisbin:
I have to make one more change: renaming branded -> commercial means I need to flip the booleans. Personal-branded -> Business-commercial

templates/plans/features.hamlet
```diff
@@ -1,8 +1,9 @@
 <ul>
-  <li>#{planSiteQuota plan} Site(s)
-  <li>#{planCommentQuota plan} Comments per month
-
-  $if planBranded plan
+  $if planCommercial plan
+    <li>#{planSiteQuota plan} commerical Site
```
cpytel: typo in "commerical"

templates/plans/features.hamlet
```diff
@@ -1,8 +1,9 @@
...
+    <li>No branding
+  $else
+    <li>#{planSiteQuota plan} non-commerical Site
```
cpytel: Another tip in "commerical"

pbrisbin: Tab-complete sucks when the only value in scope is already typod

migrations/create-plan.psql
```diff
@@ -5,6 +5,6 @@ CREATE TABLE IF NOT EXISTS plan (
   price                 INT8 NOT NULL,
   site_quota            INT8 NOT NULL,
   comment_quota         INT8 NOT NULL,
-  branded               BOOLEAN NOT NULL,
+  commercial            BOOLEAN NOT NULL,
```
gabebw: How do migrations get run in Yesod? Editing old ones is A-OK?

pbrisbin: Migrations don't run unless you run them with `bin/migrate-deploy`, so this would never get automatically re-run. I've changed it because it's used by `bin/setup-db` to bootstrap a new development instance.

gabebw:
I think this looks good.

PR #256, opened 2015-04-17
Call withEnv in TestImport.withApp

pbrisbin:
In e7cc415f, loadEnv was moved out of makeFoundation and placed earlier in the
application boot process. This unintentionally meant it was no longer invoked
during tests was part of withApp. This was OK on CirclCI where the environment
variables are explicitly set, but breaks tests locally where they are set in
.env.

### DIFF
--- a/Application.hs
+++ b/Application.hs
@@ -4,6 +4,7 @@ module Application
     , appMain
     , develMain
     , makeFoundation
+    , withEnv
     -- * for DevelMain
     , getApplicationRepl
     , shutdownApp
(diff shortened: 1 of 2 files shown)


Application.hs
```diff
@@ -178,7 +179,7 @@ appMain = do
     -- Run the application with Warp
     runSettings (warpSettings foundation) app
 
-withEnv :: IO AppSettings -> IO AppSettings
+withEnv :: IO a -> IO a
 withEnv = (loadEnv >>)
```
pbrisbin: I'll probably move this to `LoadEnv` at some point.

jferris:
Looks good.

PR #259, opened 2015-04-17
Fix site migrations

pbrisbin:
These were never updated when we changed the Site model. They are used in
bin/setup-db and therefore need to be maintained.


gabebw:
works, :+1:

PR #260, opened 2015-04-17
Fetch user data asynchronously

### LINKED ISSUE #170: Synchronous XMLHttpRequest on the main thread for /user
> Synchronous XMLHttpRequest on the main thread is deprecated because of its detrimental effects to the end user's experience. For more help, check http://xhr.spec.whatwg.org/.

pbrisbin:
- Make all calls to getUser asynchronous
- Fire the first attempt right when Carnival loads
- Expect Carnival.user to be present when the form is rendered, if not
 launch the login window

Resolves #170, #248

### DIFF
--- a/templates/embed/Carnival.coffee
+++ b/templates/embed/Carnival.coffee
@@ -46,27 +46,32 @@ class Carnival
     Carnival.user.gravatar_url
 
   @isLoggedIn: ->
-    @getUser()
     if @user?
       return true
     else
       return false
 
-  @getUser: ->
-    request = new XMLHttpRequest
-    request.withCredentials = true
-    request.open('GET', '@{UserR}', false)
-    request.setRequestHeader('Accept', 'application/json')
-    request.send()
-    if request.status is 200
-      @user = JSON.parse(request.responseText).user
+  @getUser: (callback) ->
+    if @user?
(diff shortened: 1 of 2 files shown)


jferris:
Do we need to fire a request right when the script loads? That will increase the load on the server pretty dramatically, as most requests won't involve making comments.

pbrisbin:
Do you think moving it to when the sidebar is opened will help enough (assuming it works)? That would still be more than is strictly required.

pbrisbin:
Seems to still work moving it there. I could get close, but never quick enough to trigger an unnecessary login.

templates/embed/Carnival.coffee
```diff
@@ -46,27 +46,32 @@ class Carnival
...
+
+      request.send()
 
   @hasLoggedIn: (event) =>
     if event.origin != '%{root}'
       return
     @loginWindow.close()
-    @getUser()
-    document.dispatchEvent(new CustomEvent("hasLoggedIn", bubbles: true))
+    @getUser () =>
```
jferris: Not really important, but you can use `->` here, because the function doesn't use any instance state.

jferris:
Looks good to me.

PR #263, opened 2015-04-17
Add bourbon, neat and bitters. Refactor sass organization

smharley:
Mostly just adding in bourbon, neat and bitters. 
Splitting up the former `base.scss` file into `components`

Pages should still look the same:
image
image
image


joshuaogle:
LGTM :+1:

Issue #268, opened 2015-04-28
Syntax error on docs page


calebthompson:
![](http://caleb.click/anlY/Image%202015-04-28%20at%201.18.12%20PM.png)

jferris:
Fixed: https://carnivalapp.io/docs

PR #269, opened 2015-04-28
Fix syntax error in JavaScript example

jferris:
There was an extra semicolon in the addEventListener snippet.


jferris:
Fixes #268.

Issue #257, opened 2015-04-17
When comments run longer than the article, they bleed into footer


pbrisbin:
![shot](https://cloud.githubusercontent.com/assets/50812/7206813/173484ac-e503-11e4-8e9c-db2d201b583c.png)

gabebw:
`overflow: scroll`?

pbrisbin:
It's a bit more complicated than that. These threads are contained in a `.carnival` div of no size and absolutely positioned to line up with the paragraphs. I think we would need to get that div to be the same height as the article, before we can think about how to position/constraint the threads within it.

smharley:
`max-height` might also be a band-aid

jferris:
The proposed `max-height` band-aid has been deployed, but it would be nice to come up with a more robust fix.

pbrisbin:
Wild idea:

- Let `d` be the distance from the top of the thread from the top of the article
- Let `h` be the height of the article element
- In JS, set a `max-height` of `h - d` for each thread element
- Profit?

PR #270, opened 2015-04-30
Use CSS to stop comments going below the fold on the home page

smharley:
Not the ideal solution, but could help with https://github.com/thoughtbot/carnival/issues/257 until we can calculate the correct height with js or something. 

comments


jferris:
I think this looks good to merge.

jferris:
Merged as b1f33eb and will be deployed shortly.

PR #272, opened 2015-05-01
Add styles for plans page

smharley:
image


jferris:
Looks good!

I think we should get some feedback from @thoughtbot/design on the CSS and visuals.

smharley:
Agreed! Just pushed up a couple fixes for the stripe button, will get feedback shortly :racehorse:

kylefiedler:
Why is the call to action for enterprise in a different spot than the other plans? Should we play down the free version more?

Code looks good.

smharley:
@kylefiedler mostly because there isn't a price, you have to contact us for both pricing and subscribing. Do you think it would be better to have "Contact us" in both locations? Something else?

donokuda:
@jferris Are comments limited for Business and Personal plans? I see "Unlimited Comments" in the enterprise section but no mention about limited comments elsewhere.

kylefiedler:
@smharley What about `Prices Vary` and have the button at the bottom say `Contact us for Pricing`? It just feels less important compared to the others where as I would assume it's second most important.

ldamon:
Does "no branding" mean "you can use your own branding"?
If so, I might say, "no carnival branding" or "your own branding" or "customized styling".

smharley:
:+1: good feedback everyone, thanks!

jferris:
> Are comments limited for Business and Personal plans?

No, not right now. We talked about doing that, but we didn't apply comment quotas for the initial release.

smharley:
Updated design: @donokuda @kylefiedler 
image

donokuda:
LGTM

PR #273, opened 2015-05-01
Delete demo comments after one week

jferris:
Comments on the demo site are for the purpose of trying out the site. We
expect a lot of "test and "hello" type of comments, and the page will
look strange with hundreds of old test comments on it.

This commit adds a subcommand to the `carnival` executable to delete
stale comments from the demo site.

### DIFF
--- a/Application.hs
+++ b/Application.hs
@@ -3,6 +3,7 @@ module Application
     ( getApplicationDev
     , appMain
     , develMain
+    , taskMain
     , makeFoundation
     , withEnv
     -- * for DevelMain
@@ -28,6 +29,7 @@ import Network.Wai.Middleware.RequestLogger (Destination (Logger),
                                              mkRequestLogger, outputFormat)
 import System.Log.FastLogger                (defaultBufSize, newStdoutLoggerSet,
                                              toLogStr)
+import Tasks                                (parseTask)
 
 import LoadEnv
 import System.Environment (getEnv)
@@ -178,6 +180,9 @@ appMain = do
 
     -- Run the application with Warp
     runSettings (warpSettings foundation) app
+
+taskMain :: IO ()
+taskMain = parseTask >>= handler
 
 withEnv :: IO a -> IO a
 withEnv = (loadEnv >>)
--- a/carnival.cabal
+++ b/carnival.cabal
@@ -40,6 +40,7 @@ library
                      Handler.Purchase
                      Handler.Plan
                      Handler.Cancel
+                     Handler.DeleteStaleComments
                      Helper.Request
                      Helper.Validation
                      SendMail
@@ -49,6 +50,7 @@ library
                      Language
                      Network.Mail.SendGrid
                      Network.Mail.RecipientOverride
+                     Tasks
 
     if flag(dev) || flag(library-only)
         cpp-options:   -DDEVELOPMENT
@@ -112,6 +114,7 @@ library
                  , containers
                  , vector
                  , time
+                 , optparse-applicative
 
                  , SHA                           >= 1.6.4      && < 1.7
                  , blaze-html
@@ -141,6 +144,13 @@ executable         carnival
         Buildable: False
 
     main-is:           main.hs
+    hs-source-dirs:    app
+    build-depends:     base, carnival
+
+    ghc-options:       -threaded -O2 -rtsopts -with-rtsopts=-N
+
+executable         carnival-task
+    main-is:           TaskMain.hs
     hs-source-dirs:    app
     build-depends:     base, carnival
 
--- a/test/Model/CommentSpec.hs
+++ b/test/Model/CommentSpec.hs
@@ -1,0 +1,46 @@
+module Model.CommentSpec
+    ( main
+    , spec
+    ) where
+
+import TestImport
+import Model.Comment
+import Data.Time.Clock (addUTCTime)
+
+main :: IO ()
+main = hspec spec
+
+spec :: Spec
+spec = withApp $
+    describe "deleteStaleComments" $
+        it "deletes comments older than a week for the given site" $ do
+            siteId <- createSiteWithUrl "http://example.com/site"
+            otherSiteId <- createSiteWithUrl "http://example.com/otherSite"
+            Entity u _ <- runDB $ createUser "example"
+            runDB $ do
+                setCreatedAgeInDays 6 =<<
+                    createComment u siteId "x" "x" "recent"
+                setCreatedAgeInDays 8 =<<
+                    createComment u siteId "x" "abc" "old"
+                setCreatedAgeInDays 8 =<<
+                    createComment u otherSiteId "x" "x" "other site"
+
+            runDB $ deleteStaleComments siteId
+
+            result <- runDB
+                $ map (commentBody . entityVal)
+                <$> selectList [] [Desc CommentBody]
+
+            result `shouldBe` ["recent", "other site"]
+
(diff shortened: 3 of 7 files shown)


jferris:
cc @pbrisbin @thoughtbot/haskell

Application.hs
```diff
@@ -162,22 +165,18 @@ develMain = develMainHelper getApplicationDev
...
-    -- Generate a WAI Application from the foundation
     app <- makeApplication foundation
 
-    -- Run the application with Warp
-    runSettings (warpSettings foundation) app
+    command <- parseCommand
+
+    case command of
+        RunServer -> runSettings (warpSettings foundation) app
+        DeleteStaleComments -> handler $ do
```
jferris: I'm guessing there's a better place to organize these. I wondered if it would be better to make the delete commands task an actual handler module, and then import/use it here.

carnival.cabal
```diff
@@ -137,9 +139,6 @@ library
                  , yesod-text-markdown
 
 executable         carnival
-    if flag(library-only)
```
jferris: This felt strange to do, but you can't build the executable in development mode without this change.

I also thought about adding a new executable (`carnival-task` or something), but it seemed wasteful to build two executables with all the same dependencies.

test/Model/CommentSpec.hs
```diff
@@ -0,0 +1,46 @@
...
+spec :: Spec
+spec = withApp $
+    describe "deleteStaleComments" $
+        it "deletes comments older than a week for the given site" $ do
+            siteId <- createSiteWithUrl "http://example.com/site"
+            otherSiteId <- createSiteWithUrl "http://example.com/otherSite"
+            Entity u _ <- runDB $ createUser "example"
+            runDB $ do
+                setCreatedAgeInDays 6 =<<
+                    createComment u siteId "x" "x" "recent"
```
jferris: We need to figure out a better test data solution...

Application.hs
```diff
@@ -162,22 +165,18 @@ develMain = develMainHelper getApplicationDev
...
-    -- Get the settings from all relevant sources
-    settings <- withEnv $ loadAppSettingsArgs
-        -- fall back to compile-time values, set to [] to require values at runtime
-        [configSettingsYmlValue]
-
-        -- allow environment variables to override
-        useEnv
-
-    -- Generate the foundation from the settings
+    settings <- getAppSettings
```
jferris: Using `loadAppSettingsArgs` makes Yesod swallow the args while figuring out the environment. I'm not sure if we're using these arguments in production.

Command.hs
```diff
@@ -0,0 +1,26 @@
...
+import Import hiding ((<>))
+import Options.Applicative
+
+data Command = RunServer | DeleteStaleComments
+
+parseCommand :: IO Command
+parseCommand = execParser $ info (helper <*> commandParser) fullDesc
+
+commandParser :: Parser Command
+commandParser = subparser
```
gabebw: Would this be easier to read as `subparser serverCommand <> deleteStaleCommentsCommand <|> pure RunServer`

Application.hs
pbrisbin: The default behavior (provided by `loadAppSettingsArgs` is to accept a filename. A feature I use [here](https://github.com/thoughtbot/carnival/pull/250/files#diff-0405cbeab00d5d74e15d255c36022ee4R8). It might also be used implicitly by `yesod test`, I'm not sure. Either way I think we need to be careful about removing it.

Command.hs
```diff
@@ -0,0 +1,26 @@
...
+    ( command "server"
+        ( info
+            (pure RunServer)
+            (progDesc "Run the Carnival HTTP server (default)")
+        )
+    <> command "delete_stale_comments"
```
pbrisbin: I've never seen underscored subcommands. What about hyphens? (e.g. `git rev-parse`)

carnival.cabal
pbrisbin: I like the idea of separate executables. I'm not sure what's wasted in that case, though managing the dependency lists may be a hassle.

test/Model/CommentSpec.hs
pbrisbin: I've slowly extracting a generict `buildX` out of all the `createX` functions we have currently. This could be a good oppurtunity to make `insert $ buildComment { commentCreate = x, commentBody = "recent" }` work...

Application.hs
pbrisbin: A separate executable would also solve this issue.

jferris: I don't think a separate executable solves the issue, because we'll still need to load the app settings in the other executable. Loading the settings differently for different executables seems confusing to me.

carnival.cabal
jferris: It's slower to build, it increases the size of the slug, and it complicates the cabal file.

I can do a version which uses a separate executable to see what the actual differences are.

jferris:
@pbrisbin I did a pass adding another executable instead of adding a subcommand: https://github.com/thoughtbot/carnival/commit/8728a62e17f833e59c58620d575c4fc8d9372aba

Some observations:

* The cabal changes weren't bad at all
* The build time did not noticeably increase (it just links another executable)
* The slug size will jump from 13M to 22M
* Halcyon correctly copied the additional binary into the `bin/` directory in the slug

Given the small size of our starting slug, I think having two executables is okay. If we continue to add tasks, though, I don't think we'll want to add an executable for each one.

I deployed the separate-executable version to staging, and it seems to pick up the application config correctly. For example, it connected to the staging database and correctly deleted old comments from the demo site on staging.

pbrisbin:
That all sounds great. To clarify, I wouldn't advocate a separate executable for each task, I'd advocate two executables: one to run the app, and one (with sub-commands) to run any tasks.

jferris:
> To clarify, I wouldn't advocate a separate executable for each task, I'd advocate two executables: one to run the app, and one (with sub-commands) to run any tasks.

Yeah, I wanted to see what it would be like to just add a new executable for each one, Mike Burns Style.

I'll mess with that other commit to put in a framework for adding subcommands.

pbrisbin:
Ha. I would use mbx (Mike Burns Unix).

jferris:
Okay, I pushed another commit. I'm pretty happy with how this is shaping up.

There's now a `carnival-task` executable which accepts subcommands. I added a `TaskMain` and `taskParser`, which uses `optparse-applicative` to transform arguments into a `Handler ()`.

I think it will be easy to add tasks in the future, and feels pretty normal, like adding new Yesod routes. You just add a parser step instead of a route and then write a `Handler` as normal.

pbrisbin:
LGTM -- having tasks be `Handler` is strange to  me, but I get why. Do be aware that `handler` does eventually call a function with "unsafe" in its name -- though, I think we know what we're doing here.

jferris:
> LGTM -- having tasks be Handler is strange to me, but I get why. Do be aware that handler does eventually call a function with "unsafe" in its name -- though, I think we know what we're doing here.

I played around for a while trying to find less specific type than `Handler`. The biggest issue seems to be that methods based around getting the app settings like [`getYesod`](http://haddocks.fpcomplete.com/fp/7.8/20140916-162/yesod-core/Yesod-Core-Handler.html#v:getYesod) only work within handlers.

It seems like it would be possible to have a type that represents an action within the app settings and with a database connection, but without worrying about requests/responses. However, I think that's out of the scope of this pull request.

I'm going to merge this as-is, but I'd be interested in exploring the `Handler` type more during the next project night.

PR #280, opened 2015-06-05
Use CSS transition for smoother commenting

dgalarza:
When going to comment the way an article's content is moved to the left is a bit jarring and almost feels unintentional. This PR adds a CSS transition to smooth things out providing a bitter experience.

My original thought was to remove the move all together, however this has larger implications as we would have to deal with different screen sizes. I think this provides a good middle ground with a smoother transition. The gif below doesn't really do it a lot of justice as the transition seems a big jagged there.

carnival-transition

### DIFF
--- a/templates/embed.lucius
+++ b/templates/embed.lucius
@@ -1,3 +1,12 @@
+@keyframes carnival-open {
+  0% { left: 0; }
+}
+
+@keyframes carnival-close {
+  0% { left: -150px; }
+  100% { left: 0; }
+}
+
 .carnival {
   display: none;
   position: absolute;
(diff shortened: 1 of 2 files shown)


cpytel:
I'm glad you got this working. I think the structure of the markup changed since I originally tried and couldn't get it to smoothly transition.

That being said, the strategy for carnival is that we're Trying not to provide the styles for what happens when commenting in carnival itself, rather it just puts a `commenting` class on the article when it's being commented on and the consuming blog can do that however it wants. There should already be styles in the blog for this.
Is it possible to put this change in those styles?

dgalarza:
@cpytel looks like someone added a transition for this on robots.thoughtbot.com itself, so it's no longer a jarring experience there. Do you think this is something we would want to be provided by default by Carnival though?

cpytel:
I don't have a strong opinion. It probably makes sense for this to be the blog's choice.

dgalarza:
> It probably makes sense for this to be the blog's choice.

After taking some time to pair with @tysongach this seems like the best solution at this time.

PR #283, opened 2015-11-11
Allow GitHub authentication without a name defined

### LINKED ISSUE #282: 'Authentication Error' with github sign-in
Steps to reproduce:

1. Use a github account that hasn't authorized Carnival, or revoke the Carnival access token in Github > Settings > Applications.
2. Visit https://carnivalapp.io/auth/login?dest=https://carnivalapp.io/
3. Select 'Sign in with Github'
4. Select 'Authorize Application'
5. Observe redirect to https://carnivalapp.io/auth/login with message "Authorization Error".

Expected behavior:
Some form of logging into Carnival where I could leave comments on Carnival enabled blogs.

pbrisbin:
Use Alternative to grab the profile name from the login key if/when the name
key is missing. (This required an orphan Error instance on Text.) Profiles
missing a name are possible, while login should always be present[1].

Fixes #282

1: https://github.com/thoughtbot/yesod-auth-oauth2/blob/master/Yesod/Auth/OAuth2/Github.hs#L97

I included a commit that adds stack support. Note that this is not fully-baked;
specifically, I can't get `stack test` or `yesod test` to work yet, but it does
work well enough for me to test-compile the project and run individual spec
files. If you'd prefer I keep that commit out, let me know.

/cc @jferris

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -15,3 +15,5 @@ cabal.sandbox.config
 .env
 !migrations/*.sql
 .sass-cache/
+.stack-work
+tmp
--- a/Data/Text/Instances.hs
+++ b/Data/Text/Instances.hs
@@ -1,0 +1,11 @@
+{-# OPTIONS_GHC -fno-warn-orphans #-}
+module Data.Text.Instances where
+
+import Control.Monad.Error (Error(..))
+import Data.Text (Text)
+
+import qualified Data.Text as T
+
+-- Required to use @(<|>)@ with @Either Text a@ values
+instance Error Text where
+    strMsg = T.pack
--- a/Model/User.hs
+++ b/Model/User.hs
@@ -15,6 +15,7 @@ import Import.NoFoundation
 
 import Data.Aeson
 import Data.Digest.Pure.SHA (hmacSha256)
+import Data.Text.Instances ()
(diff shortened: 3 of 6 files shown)


jferris:
This looks good to me. I don't object to the Stack changes.

While figuring out the `Either`/Text`/Error` changes, it occurred to me that maybe the whole user-to-creds thing should be in the `Error` monad instead of using `Either Text a` repeatedly. Thoughts?

pbrisbin:
Yes, we should definitely be using `ExceptT` (`ErrorT` is deprecated in favor of it).

If that were the case, we shouldn't need the orphan `Error` instance on `Text` since that constraint is `(Functor m, Monad m, Monoid e) => Alternative (ExceptT e m)` and `Text` is a `Monoid`.

I made the decision to go this way because it was a relatively quick fix for a pretty big bug and elected to save the larger `ExceptT` discussion for later -- I'm also open to exploring `ExceptT` now though, as saving it for later would probably result in me not getting to it for some time.

jferris:
I think going for the smaller fix is cool, particularly since this is a bug in production.

I might take a shot at the `ExceptT` refactoring on Friday.

PR #284, opened 2015-11-20
Fix yesod devel command

### LINKED ISSUE #279: yesod devel fails on master
We added another executable for running scheduled tasks, and it seems to have broken `yesod devel`. The error message is the same as the one reported in https://github.com/yesodweb/yesod/issues/864.

jferris:
This command broke when we added the `carnival-task` executable. Running
`yesod-devel` before this commit prints this error:

    Starting development server...
    <command line>: cannot satisfy -package-id main-inplace
        (use -v for more information)
    Exit code: ExitFailure 1

Excluding the executable in library-only mode (used by `yesod devel`)
fixes the error. After this commmit, you'll need to reconfigure and
rebuild if you want to run `carnival-task` in development. Running
`yesod-devel` again will automatically reconfigure in library-only mode.

Resolves #279.

### DIFF
--- a/carnival.cabal
+++ b/carnival.cabal
@@ -151,6 +151,9 @@ executable         carnival
     ghc-options:       -threaded -O2 -rtsopts -with-rtsopts=-N
 
 executable         carnival-task
+    if flag(library-only)
+        Buildable: False
+
     main-is:           TaskMain.hs
     hs-source-dirs:    app
     build-depends:     base, carnival


jferris:
cc @thoughtbot/haskell @pbrisbin

pbrisbin:
LGTM -- makes sense that the build configuration now matches the existing executable more too.

jferris:
> makes sense that the build configuration now matches the existing executable more too.

Yeah. I originally excluded that intentionally, because you needed to manually rebuild in order to do `cabal run carnival-task`. In retrospect, it's kind of obvious you can't have to build configurations active at the same time, so this was a silly thing to do.

PR #286, opened 2016-01-29
Use Docker

jferris:
* Adds Dockerfile, docker-compose.yml
* Uses `heroku docker:release` to deploy to staging
* Replaces Halcyon with Stack
* Updates Stack resolver to use stack-supported yesod-bin


Dockerfile
```diff
@@ -0,0 +1,29 @@
...
+  && apt-get update \
+  && apt-get install -y \
+    coffeescript \
+    postgresql-client \
+    stack \
+  && apt-get clean \
+  && rm -rf /var/lib/apt/lists/*
+
+COPY stack.yaml .
+RUN stack setup
```
tonyd256: what about stopping here and making this a `thoughtbot/stack` docker hub image?

Dockerfile
```diff
@@ -0,0 +1,29 @@
...
+
+COPY *.cabal ./
+RUN stack install yesod-bin cabal-install
+RUN stack build --dependencies-only
+
+COPY . /app/user
+RUN stack build
```
tonyd256: Then make the reset of this a `thoughtbot/stack-app` or something that we can use for any stack app we have.

bin/deploy
```diff
@@ -11,19 +11,7 @@ set -e
...
-
-      git checkout -b deploy
-      git commit --allow-empty -m "Trigger deployment"
-      git push --force staging deploy:master
-      git checkout master
-      git branch -D deploy
-
-      heroku ps:scale web=1
-    fi
+    heroku docker:release
```
tonyd256: does this now do the slug stuff we were doing?

docker-compose.yml
```diff
@@ -0,0 +1,25 @@
+db:
+  image: postgres:9.4
+  ports:
+   - "5432"
+web:
+  build: .
```
tonyd256: if you have the `thoughtbot/stack-app` image in docker hub you can change this line to `image: thoughtbot/carnival` ... then you can add a `update-image` script that will do the updating of dependencies and pushing of the new image. The bonuses here is that when someone pulls this to work on it, they won't have to reinstall the dependencies.

pbrisbin:
WDYT about holding off on trying to use the Dockerfile for development, postgres via docker-compose etc, and (for now) just shipping what's needed for `heroku docker:release` to work?

I think the various problems you're having are related to the former, but the latter is the more valuable increment, IMO. Getting just that much in would make #285 shippable, and Halcyon completely removable.

Dockerfile
pbrisbin: The installation of `postgresql-client` and `coffeescript` are carnival-specific. Migtht be fine for a `thoughtbot/stack`, but I was hope to get make a generic `heroku-haskell` image, and I don't think that'd fly for that.

pbrisbin:
FYI: I'm in the process of shipping https://hub.docker.com/r/pbrisbin/heroku-haskell-stack/builds/ which let me reduce a Heroku Docker-based project's own Dockerfile to https://github.com/pbrisbin/alac/blob/master/Dockerfile. I'm going to start using it in other projects, with Carnival being the most complex it may be a few iterations before it can be used here (if we want to).

bin/deploy
pbrisbin: @jferris this line is probably missing an `--app carnival-staging`.

@tonyd256 Among other things https://github.com/heroku/heroku-docker/blob/master/commands/release.js#L43

IMO the fact that it requires a compose file and runs `docker-compose build web` instead of just `docker build` is bizarre. It also requires an `app.json` so it can sort out add-ons instead of only doing that *if* there's an `app.json`. A little annoying that you need the extra boilerplate, but oh well.

Dockerfile
jferris: I think it would be fine to install `postgres-client` in a generic Heroku image (or a thoughtbot image), since it's part of the core Heroku stack. It's also a fairly lightweight dependency.

I plan on extracting a more reusable Docker image from this once we get it working, but it's easier to develop everything with a simpler base image while it's still in flux.

bin/deploy
jferris: It used to work without `docker-compose` and just run `docker build`, but that was annoying if you were using docker-compose locally for development, because it wouldn't reuse the same image layers. I don't really understand why `docker build .` and `docker-compose build web` don't share layers, but this change was pretty convenient for me.

@pbrisbin Are you saying I need to ad an `app.json`? Does that replace the `Procfile`?

docker-compose.yml
jferris: Yeah, I like what you've been doing there. I'd like to do that as an additional change in a new pull request once we get the image building and deployed.

jferris:
> WDYT about holding off on trying to use the Dockerfile for development, postgres via docker-compose etc, and (for now) just shipping what's needed for heroku docker:release to work?

Yeah, that's probably a good idea. I think I can simplify a few things by doing that.

For what it's worth, I have the issue with `stack exec yesod devel` with or without Docker. I'm not sure how it's working for you, as the `stack.yml` directly specifies an incompatible version of `yesod-bin`. Maybe it will use a newer `yesod-bin` if you somehow have it installed from another project?

bin/deploy
pbrisbin: We will need an `app.json` for this command to succeed. It does not replace the `Procfile`. `app.json` is used to specify add-ons. I'm miffed in general that it doesn't take the "no `app.json` present" case as "no add-ons needed", but in our case we could actually use it to specify the add-ons Carnival uses if we wanted. I've not used this feature yet.

pbrisbin:
I'll try to dig into the stack-yesod-devel thing and verify that it's really working and why.

jferris:
I simplified this to just replace deployment instead of development. I also added some missing stuff for deploying (`--app carnival-staging`, `app.json`, and `Procfile`). This is ready for re-review.

pbrisbin:
This looks ready to try to me.

jferris:
The staging deploy was successful.

One weird thing is that we're now building for production using Docker/Stack, and building locally/on CI using Halcyon and either cabal-install or Stack (I can't even tell!).

I think it makes sense to switch everything over to Stack at the same time as production, so that we have better production/development/test parity. We can try and switch dev/test over to Docker later if we think that's beneficial.

Objections?

pbrisbin:
It sounds like you're advocating merging this, #285, and the removal of anything halcyon basically at once. If so :+1:

Procfile
```diff
@@ -0,0 +1 @@
+web: ./carnival
```
pbrisbin: Is this missing a worker process?

jferris: Carnival doesn't have background jobs....right?

pbrisbin: You added one to periodically scrub the example comments on the homepage, I think.

https://github.com/thoughtbot/carnival/blob/master/carnival.cabal#L153

jferris: Ah. That isn't a worker process; it's a binary (`carnival-task`) which gets run daily by Heroku scheduler.

pbrisbin: Oh cool. I didn't know (or was told and didn't internalize) about the scheduler thing. That'd be better than what I'm doing in tee-io which is basically `forever $ run >> threadDelay`.

jferris:
This is ready for review again. I've now replaced Halcyon with Stack entirely and updated the documentation accordingly. I borrowed a bunch of code from #285, so this does just about everything except updating GHC, which I thought made sense to handle separately.

.dockerignore
```diff
@@ -0,0 +1,10 @@
+.git
+.halcyon
```
pbrisbin: Can probably ditch this, if that code's `rm`d now.

.dockerignore
```diff
@@ -0,0 +1,10 @@
...
+.stack-work
+Dockerfile
+dist
```
pbrisbin: I don't think we'll see this directory show up unless someone uses bare `cabal`.

pbrisbin:
Well done. LGTM.

.dockerignore
jferris: This shows up again for me whenever I run `stack exec yesod devel`.

PR #287, opened 2016-02-03
Ignore .env in when building slug

jferris:
If you have development settings in .env, they'll override the Heroku
configuration.

This avoids sending .env during the Docker build.


pbrisbin:
Nice catch. LGTM.

PR #288, opened 2016-07-15
Update to LTS 6.7

jferris:
The biggest changes here are from the updated Stripe library:
- There's no Stripe monad anymore, so I had to create one
- The procedure for constructing Stripe requests is very different

### DIFF
--- a/Stripe.hs
+++ b/Stripe.hs
@@ -6,14 +6,20 @@ module Stripe
 
 import Import
 
+import Control.Monad.Except (ExceptT, runExceptT, throwError)
+import Web.Stripe ((-&-))
+
 import qualified Web.Stripe as S
 import qualified Web.Stripe.Customer as S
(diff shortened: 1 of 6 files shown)


jferris:
cc @pbrisbin

Stripe.hs
```diff
@@ -25,36 +31,33 @@ subscribeToPlan :: S.TokenId
...
-    Nothing         -- ExpMonth
-    Nothing         -- ExpYear
-    Nothing         -- CVC
-    Nothing         -- CouponId
-    Nothing         -- Description
-    (Just email)    -- Email
-    Nothing         -- PlanId
-    Nothing         -- Quantity
-    Nothing         -- TrialPeriod
-    []              -- MetaData
```
pbrisbin: Good riddance ;)

pbrisbin:
LGTM

